PR #13, opened 2026-07-23
feat(account): personal account settings and workspace switching

pulkitxm:
Adds /settings/account (profile, connected accounts, passkeys, sessions), a workspace switcher that actually switches, and workspace creation at /workspaces/new.


pulkitxm:
Walked the whole flow in a real browser (Playwright, virtual WebAuthn authenticator over CDP) against the seeded dev workspace.

**Profile**

Profile

**Connected accounts, GitHub linked and Disconnect blocked because it is the only sign in method**

Connected accounts

**Passkey added from the browser, listed with device type and created date**

Passkey added

**Disconnect unlocks once the passkey backs up the provider**

Disconnect enabled

**Sessions, current device marked, revoke one or all others**

Sessions

**Create workspace, slug suggested from the name**

Create workspace

**Lands on the new workspace board with its bootstrapped team and seven states**

New workspace board

**Switcher lists both workspaces with the active one checked**

Switcher

**Switching back sets the active organization and lands on that workspace**

Switched back

**Tenant isolation: a teammate who belongs only to Noveum never sees the other workspace**

Isolation

pulkitxm:
@coderabbitai review

PR #17, opened 2026-07-23
fix(realtime): bind the socket to the organization the client is viewing

pulkitxm:
Resolve the active organization through one shared helper and let the client state which workspace its socket is for, so the page and the realtime stream can never disagree.


pulkitxm:
Playwright pass rate over five consecutive full-suite runs (`pnpm --filter @orbit/web exec playwright test`, reseeded before every run):

- before: 0/5. `issues.spec.ts` and `same-user-tabs.spec.ts` failed on their realtime assertions in 4 of the 5 runs, and `second-workspace-realtime.spec.ts` failed in all 5. The realtime log confirmed the cause: the page rendered `org_noveum_demo` while the same user's sockets reported `organizationId: cc6d740e-...`, the second workspace.
- after: 5/5, 7 tests each run.

`pnpm verify` is green: lint, comment policy, typecheck and unit tests, 8/8 tasks.

PR #19, opened 2026-07-23
fix(realtime): reject a malformed stated organization

pulkitxm:
A blank or over long organizationId query parameter fell back to the server's own choice instead of being rejected, reintroducing the workspace mismatch the parameter exists to prevent.

### DIFF
--- a/apps/realtime/src/active-organization.test.ts
+++ b/apps/realtime/src/active-organization.test.ts
@@ -111,6 +111,16 @@ describe('client stated organization', () => {
     expect(await client.waitForClose()).toBe(ORGANIZATION_FORBIDDEN_CLOSE_CODE);
   });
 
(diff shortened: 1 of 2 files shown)


pulkitxm:
### Proof the tests catch the defect

Both new cases were run against `main`'s `server.ts` first, then against the fix.

**Before** (`apps/realtime/src/server.ts` reverted to main):

```
 × rejects an empty stated organization instead of choosing one       5005ms
 × rejects a stated organization longer than an identifier can be     5004ms
 Tests  2 failed | 3 passed (5)
```

Both time out waiting for a close frame, because the connection is accepted and the server silently picks a workspace.

**After:**

```
 Tests  25 passed (25)          # apps/realtime, was 23
```

`pnpm verify`: `Tasks: 8 successful, 8 total` for lint, comment policy, typecheck, and tests.

PR #20, opened 2026-07-23
feat(import): import the Noveum workspace from Plane

pulkitxm:
Exports every accessible Plane project, work item, comment, page and embedded image into a local cache, maps it onto the Orbit schema preserving work item numbers, and reconciles the two sides.


pulkitxm:
### Reconciliation against Plane

`pnpm db:verify-plane` counts both sides and exits non zero on any difference.

```
Work items per project
  ok   Sales (SALES)                 plane   180  orbit   180
  ok   Marketing (MKT)               plane   381  orbit   381
  ok   Customer Issues (CUST)        plane     5  orbit     5
  ok   Noveum AI (NOVEUM)            plane   280  orbit   280
  ok   API.market Engineering (ENG)  plane   432  orbit   432
  ok   Noveum.ai V1 Plan (V1)        plane     0  orbit     0

Work items assigned, by person
  ok   aditi     plane   404  orbit   404
  ok   tanzeela  plane   216  orbit   216
  ok   pulkit    plane   111  orbit   111
  ok   shivam    plane    99  orbit    99
  ok   shashank  plane    98  orbit    98
  ok   koushik   plane    73  orbit    73
  ok   harkirat  plane    55  orbit    55
  ok   pragati   plane    39  orbit    39
  ok   vidushi   plane    31  orbit    31

Work item state
  ok   open       plane   308  orbit   308
  ok   completed  plane   866  orbit   866
  ok   cancelled  plane   104  orbit   104

Documents
  ok   pages  plane    57  orbit    57

Comments  plane 120  orbit 120

Every counted total matches Plane.
```

`pnpm verify`: `Tasks: 8 successful, 8 total` for lint, comment policy, typecheck and tests. Playwright: `7 passed`.

Screenshots are deliberately not attached to this pull request. This repository is public and the imported workspace is real internal work, so the only images that belong here are of demo data.

### Two projects could not be read

`API.Market SEO Sprint` and `API Market LinkedIn` return `403 You don't have permission to view this project` for the workspace API key, and the Plane MCP bot gets the same. They are listed but not readable. Adding the key's account as a project member in Plane and re-running `pnpm plane:fetch && pnpm db:import-plane` picks them up.

### Bugs this surfaced, all pre-existing

- the issue list only ever fetched the first 200 issues even though the API has paged with a cursor since the start, so ENG would have shown 200 of 432
- the dev login listed the twelve oldest users, which breaks when every imported user shares one timestamp
- a six character team key made a ten character identifier, which wrapped onto a second line in the fixed width column

pulkitxm:
All six findings fixed in f3b23f2, each with a test that fails without the fix (33 tests in @orbit/db, was 24). The page-image path had no data exercising it in this export, so the fix is covered by unit tests over inlineAssets rather than by the reconciliation totals.

PR #27, opened 2026-07-23
fix(keyboard): scope the hotkey registry and stop advertising shortcuts that do nothing

pulkitxm:
Give hotkey entries a scope and a priority so the page decides which shortcut wins rather than registration order, stop the registry eating Enter and Space on focused controls, and define every command once so a key hint cannot drift from what it does.

### DIFF
--- a/apps/web/src/components/command-palette.test.tsx
+++ b/apps/web/src/components/command-palette.test.tsx
@@ -1,0 +1,200 @@
+import { beforeEach, describe, expect, it, mock } from 'bun:test';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { type ReactNode, useState } from 'react';
+import { CommandPalette } from '@/components/command-palette.tsx';
+import { ShortcutsOverlay } from '@/components/shortcuts-overlay.tsx';
+import { HOTKEY_PRIORITY, HotkeyProvider, useHotkey } from '@/lib/keyboard/index.ts';
+import { buildNavigation } from '@/lib/navigation.ts';
+
+const push = mock();
+const setTheme = mock();
+
+mock.module('next/navigation', () => ({
+  useRouter: () => ({ push, replace: mock(), refresh: mock() }),
+  usePathname: () => '/inbox',
+}));
+
+mock.module('next-themes', () => ({
+  useTheme: () => ({ resolvedTheme: 'dark', setTheme }),
+}));
+
+const sections = buildNavigation([{ id: 'team_1', key: 'ENG', name: 'Engineering' }]);
+
+const noop = () => undefined;
+
+function Palette({
+  startOpen = false,
+  onToggleSidebar = noop,
+  onShowShortcuts = noop,
+  children,
+}: {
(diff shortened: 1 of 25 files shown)


pulkitxm:
## The reported bug, and why it was not where it looked

> "If I have to press C then it opens Create New Issue but if I go on the docs page it shows me the C button only, which should intentionally be opening a new document."

The docs page already registered `c` for New doc. It was correct. The problem was that `selectMatch` had no way to prefer it.

`IssueWorkspaceProvider` lives in the `(app)` layout, so it mounts once and stays mounted across every navigation. Its global `c` therefore sat earlier in the registry `Map` than the docs binding, which mounts and unmounts with the page. With equal step count and equal shift specificity, `beats()` returned false and the first entry won. Page context never entered the decision at all, so the behaviour depended on which page you had visited first.

Entries now carry a `scope` and a `priority`, with a named ladder (`global` 0, `surface` 1, `layer` 2), and `beats()` compares priority **before** step count. The docs binding claims the docs surface, so it wins there and only there, regardless of registration order. When the reader lacks write permission it correctly falls back to quick create.

Proven by driving the real browser: on `/docs`, pressing `c` creates "Untitled doc" and opens the editor. On an issues page the same key opens the create issue dialog.

## Two more things the registry was getting wrong

**It was eating native activation.** Global `enter` and `space` called `preventDefault` on every keydown outside an editable element. So tabbing to a button and pressing Enter navigated to the active issue instead of activating the button, and doing it with a dialog open destroyed the unsaved draft. Activation keys now return early when the event target sits inside a real control (anchor with href, button, summary, or a button, menuitem or option role), before matching or preventing anything. Modifier combinations are untouched, so `mod+enter` submit paths still work.

**Escape had several owners.** Composed into one owner per surface that pops the innermost layer first: the peek if it is open, otherwise the selection. Both owners keep `preventDefault: false` so Radix dismissal still runs.

## Stopping the UI from lying

You also said there are buttons that do nothing. `lib/navigation.ts` is now the single definition of a command: label, section, icon, binding and behaviour in one record. The sidebar tooltip, the command palette row and the `useHotkey` registration all read that same record, so `app-shell.tsx` lost six duplicate registrations and a hint cannot drift from what it does. The shortcuts overlay is generated from the live registry, so it can only show bindings that are actually registered and enabled right now.

A test walks every rendered key hint in the tree and fails if any advertised key has no registered binding.

## Verification

`bun run verify` exits 0, 639 tests, 0 failures, with the web app going from 206 to 222.

The 16 new tests were each mutation checked rather than merely written: reverting the priority comparison fails the two ordering tests, reverting the docs priority fails the regression test with `push` never called, reverting the composed escape loses the selection, and reverting the provider guard means the button never receives its click.

Two limits worth stating. The activation guard covers exactly the enumerated controls, so a future dialog that focuses only its content div rather than a control or an input would need `[role="dialog"]` added. And the filter bar escape test asserts the right behaviour but cannot discriminate priority on its own, because the popover autofocuses a cmdk input and the registry declines on editable targets; the ordering itself is proven by the `selectMatch` unit test, which fails when the comparison is removed.

PR #80, opened 2026-07-24
test(ci): run the Playwright e2e suite as a CI gate

pulkitxm:
Adds an e2e CI job that boots web + realtime on a fresh seeded database and runs the existing Playwright specs, and fixes the next-dev regression that gate caught: react-email statically requires prettier subpaths that were unresolvable from apps/web, so every auth-linked page 500'd in dev.


pulkitxm:
Addressed both review findings in 9b6c6e6: the e2e checkout now sets `persist-credentials: false`, and a bounded realtime `/health` readiness probe runs before the Playwright step so websocket flows do not start against a still-booting realtime process.

PR #95, opened 2026-08-05
fix: remove redundant port flag in dev script for Windows

### LINKED ISSUE #94: Bug: bun run dev fails on Windows due to POSIX port syntax
### Description
When running `bun run dev` on Windows, the script crashes with the following error:
`error: option '-p, --port <port>' argument '${PORT:-3000}' is invalid.`

### Cause
In `apps/web/package.json`, the `dev` script uses `${PORT:-3000}`. This POSIX shell parameter expansion is not supported by the Windows Command Prompt or PowerShell, causing the local development server to fail out of the box.

### Proposed Solution
Change the script to `"dev": "bun --env-file=../../.env next dev"`. 

Next.js defaults to port 3000 automatically and natively respects the `PORT` environment variable across all operating systems if provided. Therefore, the explicit `--port` flag is redundant and breaks Windows compatibility.

I have tested this fix on my fork and will open a PR for this shortly!

SinghAstra:
Fixes #94

Removed the `--port ${PORT:-3000}` flag from the `apps/web/package.json` dev script. This POSIX syntax breaks the dev server on Windows. Relying on Next.js's native port handling fixes the issue for Windows users while maintaining flexibility for Mac/Linux users to use the `PORT` environment variable.

<h3>Greptile Summary</h3>

This PR removes the POSIX-only port expansion from the web development command so it can run on Windows while retaining Bun-based environment loading.
- Changes `apps/web/package.json` from an explicit shell-expanded port argument to `next dev`.
- Leaves the build, start, and validation scripts unchanged.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| apps/web/package.json | Removes the shell-specific port expression from the development script without changing the remaining package scripts. |

</details>

<sub>Reviews (2): Last reviewed commit: ["Merge remote-tracking branch &#39;upstream/m..."](https://github.com/noveum/orbit/commit/aa2e62a1194361869534a8e014c02727921302da) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50505165)</sub>


imshashank (changes requested):
Thanks for this, and the diagnosis is right: `${PORT:-3000}` is POSIX parameter expansion, so on Windows `cmd` passes the literal string through to `next` and the dev server never starts.

I verified the replacement behaves identically on macOS. Running `PORT=3007 bun run --filter '@orbit/web' dev` on your branch, Next binds to the requested port:

```
▲ Next.js 16.2.11 (Turbopack)
- Local:  http://localhost:3007
```

so the override is preserved by Next's own `PORT` handling and the flag really was redundant. `bun run verify` is green with your change merged into current `main`.

Two things before this can land.

**1. Please merge `main` into your branch.** It has moved a long way since you opened this, and CI on the fork has not run the lint, typecheck, test or build jobs at all, so nothing here has actually been checked against the tree it would land on. The Vercel entry is a fork authorisation prompt rather than a build failure, so ignore that one.


One question worth answering in the PR body, since it is the reason the flag existed: `apps/realtime` uses the same `${PORT:-3100}` shape in its dev script. If you have a Windows machine to hand, does that one need the same treatment? Happy to take both in this PR or a follow up, whichever you prefer.

imshashank:
Correction to my review: ignore the question about `apps/realtime`. I checked, and its dev script is `bun --watch --env-file=../../.env src/index.ts` with the port read from `REALTIME_PORT` inside the process, so it never had the shell expansion problem. `apps/web` was the only one. Nothing extra to do.

SinghAstra:
Done! Merged main into the branch. Thanks for the review.

SinghAstra:
@coderabbitai resolve

imshashank:
Thanks for chasing this, and the diagnosis is right: `${PORT:-3000}` is POSIX parameter expansion, cmd.exe does not do it, so on Windows Next receives the literal string and the flag is worse than useless.

But dropping the flag entirely changes the port on every machine whose `.env` sets `PORT`, and that is not hypothetical: this repository's own `.env` starts with `PORT=3001`.

Here is why the two are not equivalent:

- **Today:** `${PORT:-3000}` is expanded by the *shell*, before `bun` runs. The shell does not have `PORT` in its environment, because `PORT` lives in the `.env` file that `bun --env-file` loads into the *process*. So the shell substitutes the default and Next gets `--port 3000`.
- **After this change:** `bun --env-file=../../.env` loads `PORT=3001` into the process environment, and Next reads `PORT` natively. The dev server moves to **3001**.

That breaks the documented layout. CLAUDE.md fixes web on 3000 and realtime on 3100, `apps/web/e2e/base-url.ts` builds from that, and the realtime client resolves its socket from the page origin, so a web app on 3001 and a socket host expecting 3000 disagree.

Two ways to fix it properly, either is fine:

1. **Pin it in the script and stop reading PORT for the web app:** `bun --env-file=../../.env next dev --port 3000`. Simple, cross platform, and it matches what the file already documents. `PORT` in `.env` presumably exists for something else, and this stops the web app quietly inheriting it.
2. **Keep it configurable without shell expansion,** for instance a small `bun` script that reads `process.env.WEB_PORT ?? 3000` and spawns Next, so the default is expressed in TypeScript rather than in shell syntax that only one platform understands.

I would take option 1 unless you know of a reason the web port needs to vary.

Worth saying: I only caught this because CI does not run on fork PRs here, so I checked out both your branches and ran them locally. Your other PR (#99) came through that the same way and is merged: the collision test you added is genuinely load bearing, and I confirmed it by reverting the `selectMatch` ordering fix and watching your test go red alongside the two in `binding.test.ts`. That is exactly the standard I want, so thank you.

imshashank:
I have pushed the one line change to your branch rather than leave you waiting on it, since your diagnosis was the hard part and only the remedy needed adjusting. The script now reads:

```
"dev": "bun --env-file=../../.env next dev --port 3000"
```

That keeps what you were fixing, no shell expansion anywhere, same number on every platform, while not letting the web app inherit a `PORT` that was set for something else.

I confirmed the regression by running it rather than reasoning about it, which is why I did not just merge as it stood:

```
$ bun --env-file=../../.env porttest.ts
PORT as the process sees it: 3001
```

So without the flag Next would have bound 3001 on this repository's own `.env`, and `CLAUDE.md`, `apps/web/e2e/base-url.ts` and the realtime origin all assume 3000.

Merging once the checks come back. Thanks for both of these, and for the collision test on #99, which I verified is genuinely load bearing by reverting the `selectMatch` fix and watching your test go red. That one is already merged.

PR #104, opened 2026-08-06
Fold the docs tree and put sharing on the document

imshashank:
Closes the docs half of the feedback: *"The docs UI is also pretty bad. We need to have the folders as collapsible, and we should be able to add public, private, or shareable documents as well in each one of those."*

## Folding

Every collection group and every nested page folds, backed by the existing `useSidebarDisclosure` localStorage store rather than component state, so a fold survives a reload. Keys are namespaced `docs:group:<id>` and `docs:page:<id>`.

Opening a doc reopens its group and its whole ancestor chain through a new `openAll(ids)`, while a fold the reader made by hand is never undone by a list refresh.

## Sharing

The document header carries a segmented control: Private, Workspace, Link. One click each, with `aria-pressed` reflecting the state the doc is actually in, plus a copy button once a link exists. The Link segment is hidden from anyone without `doc:publish`.

Public and link are deliberately **not** folded into one segment. A design review flagged that folding them makes public to unlisted unreachable and breaks the transition `apps/web/e2e/docs.spec.ts` exercises. Segments map to visibilities exactly, every segment dispatches its own mutation rather than returning early when already active, and the full menu behind `doc-publish` keeps all five visibilities with their existing test ids.

No API, schema or service change was needed: `shareDoc` and `docShareSchema` already cover every visibility.


imshashank:
## Adversarial review, one confirmed defect fixed

A review pass over this branch produced 5 findings; 4 were refuted on inspection, 1 survived and was reproduced with a real test.

**The confirmed one was a regression in the existing sidebar, not just the new tree.** `useSidebarDisclosure` kept its open map in per-instance state and wrote the *whole* map back to one storage key. That was harmless while the sidebar was the only consumer. This branch mounts a second consumer in the doc tree, and on `/docs/[id]` both are on screen at once, so whichever wrote last erased what the other had recorded.

Reproduced at the hook level:

```
tree.toggle('docs:page:root')   → storage {"docs:page:root": false}
sidebar.toggle('section:teams') → storage {"section:teams": false}   ← the doc fold is gone
```

The worse direction: navigating between docs runs the reveal effect, which wrote the tree's stale map and wiped every team and section fold the reader had made since the page loaded, with no interaction with the tree at all.

Fixed in a25146c by moving the map into one module-level store read through `useSyncExternalStore`, so every consumer shares one object and a write from one is immediately visible to the other. A `storage` event from another tab refreshes it too.

Three existing tests simulated a reload with `cleanup(); render(...)`, which no longer drops the store. They now reset it explicitly, which is what a page load actually does. One assertion changed meaning honestly: a group that starts closed no longer renders its content at all, where before it flashed open and then collapsed, so the test asserts absence rather than `data-state=closed`.

Refuted and deliberately not changed: search matches hidden behind a persisted fold (pre-existing on main, unchanged by this branch); the vacuous-test claim about the reopen case (mutation testing showed it does fail when persistence is removed); firing a mutation for the state a doc is already in (no defect, and the segments must not return early or public to unlisted becomes unreachable); and disclosure keys not being workspace-scoped (the hook's pre-existing design, shared with the main sidebar).

PR #105, opened 2026-08-06
Attach a file to an issue from the slash command

imshashank:
Closes: *"Make sure that we are able to add files inside a task itself by using the slash command."*

## What was actually wrong

Two separate gaps, only one of which was obvious.

**Discoverability.** The slash menu matched on the literal label "Image or file", so `/attach`, `/upload`, `/file` and `/screenshot` all found nothing. Commands now carry keywords and match a prefix of any of them.

**The create dialog.** It mounted the editor with no `onUpload` at all, because the issue has no id yet, so `runUpload` returned early and the hidden file input was never rendered. The `/image` command silently did nothing there.

## Deferred uploads

`pending-attachments.ts` holds a file until the issue exists: it mints a placeholder with `URL.createObjectURL` so an image previews for real in the dialog, then on create uploads each held file against the new issue and rewrites the placeholder into a real link.

The dialog shows how many files are waiting, and revokes the object URLs when it closes so nothing leaks.


imshashank:
## Adversarial review, four confirmed defects fixed

A review pass produced 5 findings on this branch; 1 was refuted, 4 survived refutation and **each was reproduced with a running test** before being accepted. Three of them lost a file or wrote a dead link into an issue that nothing would repair.

**1. A failed upload persisted a revoked `blob:` URL forever.** The issue is created first, with the placeholder in its body. If the S3 PUT then fails, the placeholder was left in place, the object URL was revoked anyway, and the PATCH was skipped entirely. `isSafeUrl` does not allow the blob scheme, so every reader got an image with no source and a link with no target. The partial case was worse: with two files where one failed, the PATCH *was* sent with the dead placeholder embedded, and the branch's own test asserted that as the intended result. A failure now rewrites the placeholder to the plain file name, and the description is saved whenever any placeholder changed. **The test that encoded the old behaviour has been inverted.**

**2. Create more carried the previous file's revoked placeholder into the next issue.** `setDescription('')` cannot clear the editor, because `RichTextEditor` skips a content reset for any value it has already emitted and `''` is in that set from mount. Verified directly: after resetting the controlled value to `''`, the editor DOM still held the previous content. The composer is remounted per created issue now.

**3. A file attached while the create was in flight was silently dropped and its object URL leaked.** `setPending([])` discarded entries added after submit. Only the files handed to the finalizer leave the list now.

**4. Two quick presses of the submit shortcut attached the file to the wrong issue.** The keydown path had no `isPending` guard, unlike the button. Because query-core drops the first call's per-call `onSuccess` when a second mutation replaces it, both issues were created, the file landed on the second, and the first kept the placeholder permanently. The keyboard path is guarded now.

**The harness could not have caught any of the in-flight defects**: the create mock resolved synchronously, so there was never a pending window. It can hold a create open now, and the double-submit test fails without the guard.

Refuted: the claim that the metacharacter test asserts something unviolatable (checked, the test is named and scoped correctly and does exercise real escaping).

PR #110, opened 2026-08-06
Make every listed issue clickable

imshashank:
The complaint: "all the tasks that you show in sprint planning and in the stand-up UI are just not clickable at all".

The root cause is the shared issue row. It rendered the title as a `<button>` with an onClick, so the only way to reach an issue was a plain left click. Middle click did nothing, cmd click did nothing, and the browser showed no target on hover. Every list surface in the product renders that row, so every one of them had the same dead end.

## What was audited

Every place in `apps/web` that renders an issue identifier, title or card:

| Surface | Before | Now |
| --- | --- | --- |
| Standup columns | shared row, button | link |
| Team issue list | shared row, button | link |
| My issues, list layout | shared row, button | link |
| Project issues, list layout | shared row, button | link |
| Saved view, list layout | shared row, button | link |
| Command palette search hits | list item with a select handler | link |
| Pull request list | identifier linked, title beside it as text | both inside the link |
| Peek header, Open | button calling `router.push` | link |
| Board and kanban cards | already a link | unchanged |
| Sprint issue list | already a link | unchanged |
| Sub issues on the detail page | already a link | unchanged |
| Inbox | master detail, the reading pane links out | unchanged |
| Analytics, milestones, activity feed, doc backlinks | render no issue rows | n/a |

## How it behaves

Every one of these is now a real `<a href="/issue/ENG-12">`, so middle click and cmd click work and the URL shows on hover. Where a peek already existed, the plain click still opens the peek and calls `preventDefault`, and a modified click falls through to the browser. That is the rule the board card already followed, so it moves into one shared `IssueLink` rather than living as a second copy of `isPlainClick` per call site. The card, the row, the peek, the palette and the pull request list all use it.

Nothing here adds a fifth copy of the row markup. The three list surfaces that were broken were already rendering the shared `IssueRow`, so a single fix reaches all of them.

Making the peek header a link also let the four call sites drop the identical `onOpen` callback they each passed in, along with three now unused `useRouter` calls.

## Tests

Nine new assertions, one per fixed surface, each checking that an anchor with the right href exists rather than that the title text is present:

- `IssueRow`: the title is an anchor to `/issue/ENG-7`, and a modified click does not open the peek
- `PersonWork`: every standup column row carries its own href
- `IssueList`: the virtualised team list rows carry their href
- `MyIssuesView`: both rows carry their href
- `StandupBoard`: the row and the peek both point at `/issue/ENG-2` with no scripted push
- `Board peek`: Open is a link to the issue
- `PullsView`: identifier and title sit in one link, and the pull request itself still points at the forge
- Command palette: each hit is a link, and a modified click does not route the tab behind it

Each production change was reverted in turn to confirm the tests fail without it: putting the row back to a `<button>` fails 7 tests across all five list surfaces, and reverting the palette, the pull request row and the peek fails their tests.


imshashank:
The production change is real and well covered. I reverted each change in turn in a scratch worktree: reverting `IssueRow` to a `<button>` fails exactly the seven tests claimed, reverting the peek header fails the board test, the pulls row and forge link each fail their own, both palette tests fail, and removing only the `stopPropagation()` from `dismissOnPlainClick` fails the modified-click test specifically. That is the standard I want.

One test does not hold up, and it blocks.

## The standup peek assertions test a stub defined in the same file

`standup-board.test.tsx` replaces the whole `issue-peek.tsx` module with a local stub anchor, then asserts the href and the empty `pushed` array against that stub. Both are unfalsifiable by production code. I proved it twice: replacing the real peek's `IssueLink` with a plain button left all 11 standup tests green (only `board.test.tsx` caught it), and restoring `standup-board.tsx` to the old `useRouter` plus `router.push` shape also left all 11 green, because the stub never calls `onOpen`.

The behaviour is genuinely covered by `board.test.tsx`, so the fix is small: drop the peek half of that assertion, or stop mocking `IssuePeek` for that case.

## Space no longer opens the peek when a row title has focus

The title moved from `<button>` to `<a href>`. Anchors do not activate on Space, and `activatesFocusedControl` suppresses the global hotkey whenever the target is inside `a[href]`, so the advertised Space shortcut is swallowed and the anchor ignores it. Proved with a probe: focus a row title and press Space, nothing opens on this branch; revert `issue-row.tsx` to the button and it opens. Enter still works, and a j/k user is unaffected because focus stays on the body, but tabbing to a row is also how the row marks itself active, so this is reachable. Either handle Space in `IssueLink` or drop `a[href]` from `ACTIVATION_CONTROLS` for Space only.

## Prefetch amplification worth a number before merge

`IssueLink` renders a Next `<Link>` with no `prefetch` prop, which in next 16 defaults to prefetching on viewport entry in production. Before this PR the row title was a button and the app warmed data deliberately on hover through `usePrefetchIssueDetail`. Now scrolling a virtualised list fires an RSC prefetch per row that enters the viewport, and the palette renders a Link per hit on every keystroke. The board card already did this, so it is an amplification rather than something new, but it now applies to the highest-cardinality surfaces in the product. I did not measure it against a deployment. `prefetch={false}` on the list and palette variants, keeping it for the board, is probably right.

## Smaller

- The palette nests an `<a href>` inside a cmdk `role="option"`, which ARIA forbids and which adds a tab stop per search result inside the dialog's focus trap. `tabIndex={-1}` on the palette `IssueLink` keeps middle click and cmd click without the tab stop.
- `IssueLink` declares a `title` prop no call site passes.

Everything else is clean: no server code, no new queries, no SyncActions, no schema change, no comments, no em-dashes, no `any`, no hardcoded hex, tests all under `tests/`. I merged `origin/main` locally and the full suite passes against the merged state.

imshashank:
All four points are addressed, on top of `origin/main` merged into the branch.

## The standup peek assertions

Dropped the peek half. The stub is a `div` now, so there is no anchor left in
that file to assert an href against, and the `pushed` log is gone with it. What
the test still claims, it can prove:

- the row title is a link to `/issue/ENG-2`
- `fireEvent.click` returns `false`, so the plain click stayed on the board
  instead of navigating
- the board hands the clicked issue to the peek

Mutations: removing `event.preventDefault()` from `IssueLink`'s click handler
turns that click assertion red, and changing `peeked` in `standup-board.tsx`
from `peekId` to `selectedId` fails it along with three others. The peek link
itself stays covered by `board.test.tsx` against the real component.

## Space on a focused row title

`IssueLink` now treats an unmodified space the way it treats a plain click, so
the shortcut works wherever the link appears rather than only where a list
registers the peek hotkey. I kept `a[href]` in `ACTIVATION_CONTROLS`, because
dropping it for space would let the global hotkey fire on every anchor in the
app, including ones with no peek behind them.

Two tests: the issue list focuses the second row title and presses space, and
expects the peek for `ENG-2`; the standup board does the same for `ENG-3`.
Both go red when the `onKeyDown` line is removed from `IssueLink`.

## Prefetch

Measured rather than assumed, because prefetching on viewport entry is
disabled in development (`links.js`, `onLinkVisibilityChanged`), so a dev
server would have measured zero either way. I built the branch, served it with
`next start` on a seeded workspace, and counted requests carrying
`Next-Router-Prefetch` against `/issue/*`:

| Surface | Before | After |
| --- | --- | --- |
| Team list, 19 rows on screen | 38 | 0 |
| Same list, scrolled through | 38 | 0 |
| Palette, one keystroke with 8 hits | 6 | 0 |

Two RSC requests per visible row, on a list that virtualises to thousands. So
`prefetch={false}` on the row and on the palette hit, board unchanged. The
pull request row keeps the default: it is one link per pull request on a page
capped list, and the issue is the likely next click there, same as a card.

`tests/features/issues/issue-link.test.tsx` pins all three: the row opts out,
the palette hit opts out, the card does not. Each assertion fails when its
call site changes.

## Palette tab stop

`tabIndex={-1}` on the palette link. The test focuses the search input, tabs
once, and expects focus not to land on the hit; it fails without the prop.

## Title prop

Gone.

`bun run verify` is green on the merged state.

PR #111, opened 2026-08-06
Show what a sprint actually carried: scope, points and what moved

imshashank:
Sprint numbers were wrong and the burn up hid the one thing a sprint chart exists to show: that the sprint grew.

## What was wrong

- `cycleProgress` set `scope = rows.length`, an issue count over whatever sits in the sprint right now. Estimates were invisible on a running sprint even though `outcomeOf` already counted points for a closed one.
- That count included the `canceled` category, so dropped work inflated the denominator and every percentage read low.
- The burn up had a completed series and no scope series. Work added mid sprint silently rebaselined the chart, so the sprint looked like it fell behind for no reason.
- `buildBurnUp` took the ideal line from `burnUp.length - 1` days against a sprint length rounded from the hours between the two ends. A sprint that starts at 09:00 and ends at 21:00 two weeks later drew an ideal that stopped short of the scope; one that starts at 22:00 and ends at 09:00 drew an ideal that climbed past the top of the chart.

## What it does now

`cycleProgress` replays the `cycleId` rows in `issue_activity` across the sprint window and rebuilds, for every day, which issues the sprint was carrying:

- an issue that arrives mid sprint steps the scope line up on that day and not before,
- an issue pulled out steps it back down,
- an issue that was there from the start and never moved has no activity row, so its entry comes from the sprint start,
- cancelled work leaves the scope on the day it was cancelled and is counted on its own,
- every count has a points twin, and an issue with no estimate contributes zero rather than a `NaN`.

`CycleProgress` gains `canceled`, `estimated`, `points`, `changes` (added and removed, in issues and points) and a wider `BurnUpPoint` carrying `scope`, `scopePoints` and `completedPoints`.

The board draws the scope line next to the completed area, shows the points total once anything in the sprint is estimated, and says what came in, what went out and what was cancelled. The per assignee breakdown now also leaves cancelled work out, so it agrees with the scope above it. `cycle_progress` over MCP returns all of it.

## Schema: apply this to production before merging

**This PR adds one index that production does not have yet, and nothing in the platform will create it.** Migrations here are applied by hand against the target database, so `issue_activity_cycle_moves_idx` has to be applied to the production database as part of landing this.

It is additive: one partial index, `(organization_id, created_at) where field = 'cycleId'`, on `issue_activity`. No table, column, constraint or existing index is touched. Correctness does not depend on it. Speed does: without it the replay sequentially scans the busiest table in the app.

The exact statement to run lives in `packages/db/catchup/sprint-scope-index-catchup.sql`, in the same idempotent, single transaction shape as `production-schema-catchup.sql`:

```sql
begin;

create index if not exists issue_activity_cycle_moves_idx
  on public.issue_activity using btree (organization_id, created_at)
  where field = 'cycleId';

commit;
```

Paste it into the Supabase SQL editor, or run

```
psql "$DATABASE_URL" -v ON_ERROR_STOP=1 -f packages/db/catchup/sprint-scope-index-catchup.sql
```

on the direct connection on port 5432, not the pooler on 6543. Building it takes a share lock that holds writes to `issue_activity` while it runs; the file also carries the `create index concurrently` form to run on its own through psql if that pause is not acceptable. It is applied to the local dev and test databases already.

## Tests

- `packages/core/tests/work/cycle-service.test.ts` against real Postgres: scope steps up on the day work was added and not before, steps down on the day work was pulled out, an issue filed straight into a running sprint counts from the day it was filed rather than from day one, cancelled work stays in the scope until the day it was cancelled and is counted separately, work cancelled with no recorded time is out from the first day, points add up from estimates, points arriving mid sprint land in `changes.addedPoints`, and a missing estimate counts as zero.
- `apps/web/tests/features/cycles/burn-up.test.ts`: the sprint length in days, the elapsed clamp, the scope series, the points metric, the axis staying above a scope that peaked higher than the sprint ended on, and the ideal line landing exactly on the scope for a finished sprint instead of over or under it.
- `apps/web/tests/features/cycles/cycle-analytics.test.tsx`: the drawn paths, read back out of the chart geometry into values, not the legend labels. The scope line, the completed line, the switch to points once the sprint is estimated, the ideal pace, and the whole scope line staying inside the chart when the sprint shed work late. `apps/web/tests/features/cycles/data.test.ts` covers the per assignee tally.
- `packages/db/tests/schema/index.test.ts` for the new index.

Every one of these was watched failing before the fix, and each fix was re-broken afterwards to confirm the test catches it: the scope series drawn from the completed values, the burn up input emptied, the metric pinned to `issues`, the axis dropped off the scope series, the ideal paced off the issue count, the `createdAt` guard in `inCycleAt` deleted, `droppedBy` made to drop cancelled work across the whole sprint, and `changes.addedPoints` left at zero.

`ORBIT_TEST_LANE=pr111fix bun run verify` exits 0.


imshashank:
The service-layer replay in `cycle-service.ts` is genuinely good work, and most of it is genuinely protected: I broke 23 pieces of it and 18 mutations were caught by the new Postgres tests. The activity replay is the right approach and the daily membership reconstruction is careful.

What blocks it is that the description says "each fix was re-broken afterwards to confirm the test catches it", and that is not true for six pieces, four of which are headline claims.

## Blocking: the whole UI half is untested

`cycle-analytics.test.tsx` only asserts that three legend labels render. So:

- replacing the scope series values with the completed values: 31 web tests green
- replacing the entire burn-up input with an empty array: 31 web tests green
- pinning the chart metric to `'issues'`: 31 web tests green

That last one is the exact regression this PR exists to fix. "Estimates were invisible on a running sprint" is the headline, and nothing fails when the points metric is disabled. A test that asserts a legend label exists proves the legend renders, not that the chart is drawing the right numbers, so please assert on the plotted series.

## Blocking: two server guards that implement stated claims are uncovered

`inCycleAt`'s `createdAt >= cutoff` guard is the only thing that makes an issue created straight into a running sprint step the scope up on its own day rather than on day one, because `createIssue` never writes a `cycleId` activity row. Deleting the guard leaves all 37 core tests green.

`droppedBy` can be made to drop cancelled work retroactively across the whole sprint, rather than on the day it was cancelled, with nothing failing. That is the difference between a scope line that tells the truth and one that rewrites history.

Also uncovered: `max` no longer covering the scope series, which clamps the scope line to the top of the chart; and `changes.addedPoints`, which is never asserted anywhere.

## Not blocking, and I agree with the calls you made

Leaving `outcomeOf` counting cancelled work in the stored snapshot is right: that snapshot is the recorded outcome at close and changing it would rewrite what existing rows claim. The note about `completeCycle` rolling issues over without an activity row is a real limitation and worth carrying forward as its own task rather than expanding this PR.

On the index: applying it additively rather than running `drizzle push` was the correct instinct given the shared database. Please say in the description that it needs applying to production, and I will handle it the same way I handled the doc-comment anchor column.

No authorization problem: `getCycle` gates on `issue:read` plus organization plus `assertInTeam`, no new SyncAction is published, and the new org filters are redundant but correct.

imshashank:
All three blockers are closed on `feat/sprint-scope-and-points`, with `origin/main` merged in first and `bun run verify` green against the merged state.

## B1, the history of a closed sprint rewriting itself

`cycleMembershipMoves` bounded the activity replay at `lt(createdAt, windowEnd)` and `cycleIssueFacts` then loaded current members plus `alsoTouchedBy`. An issue that sat in the sprint for its whole run and was moved out after the window matched neither, so it was never loaded at all. The fix drops the upper bound and keeps the lower one: the replay already stops walking at each day's cutoff, and `initial` is still the inverse of the first recorded move, which is exactly what rewinding through the later moves gives.

The mirror case was just as wrong and is fixed by the same change: work parked in a sprint days after it closed used to be counted from day one, because with no in-window move `initial` fell back to current membership.

`changes.removed` stays 0 for a move recorded after the window. Inside the reporting window the issue really was still in the sprint, and the whole point of the fix is that a finished sprint stops changing under later edits. A removal recorded before the sprint ended is still reported, covered by its own test.

Three tests, each watched failing first:

- `keeps the days an issue sat in the sprint when it is pulled out after the sprint ends`: an issue in the sprint for the full run, moved out with the activity row dated five days past `endsAt`. RED before the fix with the exact reported symptom, `[1 x 15]` expected and `[0 x 15]` received. Restoring `lt(createdAt, windowEnd)` afterwards puts it back to RED.
- `leaves work parked in the sprint after it ended out of every day it ran`: also RED with the bound restored, `[1 x 15]` against `[2 x 15]`.
- `reports work pulled out before the sprint ended as removed, and stops counting it that day`: the guard that the in-window path did not regress, `scopePoints` stepping 7 to 2 on the day of the move.

The mutation you called out is dead: `if (inCycleAt(entry, end))` to `if (entry.issue.member)` in `scopeChanges` now fails the first of those, because current membership says removed while membership at the window end says kept.

On cost: the replay now reads cycle moves for the organization from the sprint's start rather than only its window, so viewing an old sprint scans further along `issue_activity_cycle_moves_idx`. Any correct reconstruction has to read those later rows, since a move after the window is what tells you the state at the window, and narrowing the scan would need a different index rather than a different query. Noted rather than hidden.

## B2, the index test only checking a name

`packages/db/tests/schema/index.test.ts` now asserts the columns and the emitted predicate, `"issue_activity"."field" = 'cycleId'`, and a second test holds the catchup sql to the same definition so the file applied by hand cannot drift from the schema. Every mutation goes RED: predicate to `field = 'stateId'` fails both, dropping `created_at` from the index columns fails both, changing the predicate in the executable statement of the catchup file fails the second, and changing it in the documented `concurrently` variant fails it too.

## B3, the MCP progress assertions being type checks

`cycle_progress` is now asserted against a fixture built entirely through the tools: a new team, its sprint moved back so it opened three days ago, two sized issues carried in, one finished, then the other pulled out. It asserts `points` `{scope: 8, started: 0, completed: 5}`, `changes` `{added: 2, addedPoints: 8, removed: 0, removedPoints: 0}`, the scope series `[0, 0, 0, 2]` and the points series `[0, 0, 0, 8]`, then after the pull `changes.removed` 1, `removedPoints` 3 and the series `[0, 0, 0, 1]`. Your mutation, returning points `{0,0,0}`, changes `{0,0,0,0}` and every `burnUp[].scope` as 0 from the handler, fails it.

The old `expect.any(Number)` block is gone; the test it lived in keeps its own subject, moving an issue into the sprint and back out.

`bun run verify` exits 0: 146 shared, 85 db, 15 realtime-server, 12 realtime-client, 203 services, 499 core, 40 mcp-server, 36 realtime, 896 web, none failing.

The index still needs applying to the production database by hand, `packages/db/catchup/sprint-scope-index-catchup.sql`, before this ships.

PR #112, opened 2026-08-06
Make long documents fast to open and to type into

imshashank:
A 3,842 line document with 241 headings was slow to open and slower to type
into, because two things ran the whole document through `marked` plus the HTML
sanitiser on every keystroke.

## What was costing the time

1. `RichTextEditor` passed `content: toEditorHtml(renderMarkdown(value))` inside
   the `useEditor` options literal. That object is rebuilt on every render, so
   the parse ran on every keystroke even though tiptap only reads `content` when
   it creates the editor. The extension list was rebuilt the same way, and
   tiptap compares extensions by identity, so a fresh array meant
   `editor.setOptions` and a ProseMirror `updateState` on every render too.
2. `useEditorOutline` memoised on `content`, so `renderMarkdownWithHeadingIds`
   plus `extractHeadings` re-rendered the document and re-read all 241 headings
   for every character typed. `extractHeadings` builds a throwaway `<template>`
   per heading.

Separately, all the images in a document were requested the instant it opened.

## The changes

- Freeze the editor's opening HTML at mount and memoise the extension list.
- Compare a cheap signature of the heading bearing lines before rebuilding the
  outline, and keep the previous heading list when nothing changed, so the
  scroll spy also stops resubscribing and re-reading `offsetTop` for every
  heading on every keystroke.
- Throttle the scroll spy to one measurement per animation frame, and hold the
  outline memo in state rather than a ref written during render.
- Mark images `loading="lazy" decoding="async"` inside the sanitiser, and in the
  editor's image node.

## The signature has to over-trigger, never under-trigger

`extractHeadings` reads every `h1` to `h3` in the rendered document, so the
signature covers every line that can produce one: ATX headings after stripping
blockquote markers, list markers and indentation, fences, setext underlines, and
any line carrying raw HTML together with the contiguous block around it. Each
chosen line is keyed by its position, so a heading freed by a blank line closing
an HTML block above it counts as a change even though no chosen line's text
moved.

A setext underline is tested after the same prefix stripping, so `- Listed` over
`  ---` counts as the heading the renderer actually produces. A run of dashes is
told apart from a thematic break by how far it is indented: dashes at a column
the paragraph above does not reach close the list rather than underline it, so a
document full of `---` separators still does not rebuild while the block above
one is typed. A run of equals signs is never a thematic break, so it always
counts, which also covers the lazy `- Listed` over `===`.

`apps/web/tests/features/docs/outline-memo.test.ts` walks sixteen documents
through `outlineFor` and asserts the memo agrees with a full render: a heading
in a blockquote, one indented four spaces in a list item, one released from an
HTML block, a raw `<h2>`, one in a blockquote nested in a list item, each both
inserted and retitled in place, plus a setext heading retitled inside a bullet
item, an ordered item and a blockquote nested in a list item, one typed under
the last list item in the document, and one lazily continuing the item above it.

Coverage, measured rather than claimed. Restoring the signature this branch
started with turns 15 tests red against the suite as it stood before the setext
cases were added, and 17 red against the suite now. Within
`outline-memo.test.ts`, reverting the prefix stripping alone turns 7 red and
dropping the dash indentation rule alone turns 3 red. An earlier revision of
this description put the figure at 7, which understated it.

## Images

The lazy attributes are set on `img` inside the sanitiser's element handler, on
both the `HTMLRewriter` path and the linkedom path, so the serialiser escapes
the values and the second sanitise pass in `renderMarkdownWithHeadingIds` keeps
them. An earlier revision of this branch spliced them into already sanitised
HTML with a regex, which is the pattern #114 removed: the sanitiser legitimately
keeps an attribute value containing `>`, so `alt="a>`
took the injected `loading="` as the end of its own value and the browser was
handed a live `onerror`. `packages/services/tests/markdown/lazy-images.test.ts`
parses the output and asserts on the elements and the event handler attributes
present, for both renderers and both sanitiser paths.

## Numbers

`bun run bench:outline` builds a 241 heading, 34KB document, types thirty
characters into its body and reports the outline cost per keystroke. It needs
nothing but `bun install`: no `.env`, no database, no docker. On a clean
checkout of this branch:

```
document: 34396 bytes, 241 headings
outline per keystroke, rebuilt every time: 202.23ms
outline per keystroke, memoised on the heading signature: 0.18ms
```

The document size and the heading count are exact and reproduce on any machine.
The two timings are wall clock and move run to run: four consecutive runs on the
same machine gave 195.04, 202.23, 204.19 and 203.68ms rebuilt against 0.19,
0.18, 0.19 and 0.23ms memoised. Read them as three orders of magnitude, not as
fixed numbers.

That is the one measurement this branch defends. The earlier description carried
a table of paint, interactive and input latency figures from a Playwright and
CDP harness that is not in the repository; those rows are gone rather than
restated.

## Tests

Every production change here has a test that fails without it:

- `packages/services/tests/markdown/lazy-images.test.ts`: images carry
  `loading` and `decoding` through `renderMarkdown` and
  `renderMarkdownWithHeadingIds`, on both sanitiser paths, and a crafted `img`
  produces exactly one element and no event handler.
- `apps/web/tests/features/docs/outline-memo.test.ts`: the sixteen documents
  above, plus a document with a thematic break that must not rebuild while the
  block above it is typed, and one that pins the discriminator from both sides:
  `- item` over `---` produces no heading and stays out of the signature, while
  `- item` over `  ---` produces one and goes in.
- `apps/web/tests/features/docs/long-document-cost.test.tsx`: after twelve
  keystrokes the editor still holds the HTML it opened with and the same
  extension array it was created with, and an image mounted in the editor
  renders with `loading="lazy" decoding="async"`.
- `apps/web/tests/features/docs/use-editor-outline.test.tsx`: five scroll events
  produce one `requestAnimationFrame` and one measurement, the spy schedules
  again only after that frame has run, and a pending frame is cancelled on
  unmount.

The module mock that used to count parses in `long-document-cost.test.tsx` is
gone. `mock.module` is process global in Bun and that one was never restored, so
it leaked a counting stub into every test file loaded after it.


imshashank:
Not merging this one. The outline half is right and the win is real, but the image half reintroduces the stored XSS that #114 closed this morning, by the same mechanism.

## Blocking: `deferOffscreenImages` rewrites sanitized HTML with string surgery

This is the exact pattern #114 removed from `renderMarkdownWithHeadingIds`: sanitize, then re-open the output with a regex and splice attributes in. The sanitizer legitimately keeps attribute values containing `>`, because that is spec-legal HTML, so a crafted `img` breaks out of its own attribute quoting and the splice hands the browser a live element.

It reaches `bodyHtml` for comments, `descriptionHtml` for issues, `contentHtml` for the doc body, and the unauthenticated `/d/[token]` share page, all of which land in `dangerouslySetInnerHTML`. There is still no CSP.

The fix is the one #114 used: never rewrite sanitized HTML. Add `loading` and `decoding` to `ALLOWED_ATTR` in `packages/services/src/markdown/sanitize.ts` and set them on `img` inside the sanitizer's element handler, where the serializer escapes values properly.

## Blocking: the lazy-image attributes are stripped the moment this merges main

`main` now re-runs `sanitizeHtml` over the heading-id output. `loading` and `decoding` are not allowed attributes, so the second pass deletes exactly what this PR adds, on exactly the path the doc body uses. Verified on a clean merge of `origin/main` into the branch: `renderMarkdownWithHeadingIds` emits `` with both attributes gone. The "image requests on open: 240 -> 8" row is false against today's main, and the markdown suite stays green at 54 pass because the new test only exercises `renderMarkdown`, never the heading-id variant.

## Blocking: three of six production changes have no test

The description says each change was reverted to confirm its test goes red. Reverting these three leaves the full 851-test web suite green:

- `Image.configure({ HTMLAttributes: { loading, decoding } })` in `editor/extensions.ts`
- the `requestAnimationFrame` throttle in `use-editor-outline.ts`
- the `useMemo` around `editorExtensions` in `rich-text-editor.tsx`

The other changes are genuinely load bearing, so this is worth fixing rather than restating.

## `headingSignature` under-triggers, so the outline goes stale

The description says it over-triggers and is never stale. It only matches ATX headings at column 0 to 3 of a top-level line, while `extractHeadings` reads every `h1`-`h3` in the rendered document. Five reproduced cases: a heading in a blockquote, one indented four spaces in a list item, one swallowed by an HTML block, a raw `<h2>` in the body, and one in a blockquote nested in a list item.

Consequence: with `# One / > ## Inserted / ## Two`, `outlineFor` returns the identical memo, so the sidebar shows `[One, Two]` while the document has three headings. Clicking "Two" scrolls to "Inserted", because `goTo` resolves through the live DOM. A heading in a blockquote is ordinary authoring.

## The performance numbers are not in the diff

The CPU profile, the named function costs, the 175KB / 241 heading / 240 image fixture and the eight-row table are all in the description with nothing in the repository that reproduces them: no spec, no fixture, no benchmark. The call-count tests you added are a reasonable proxy and I am happy with that as the regression guard, but please either commit the harness or drop the table to the one claim that is defensible. The outline win is real, independently measured at 180.8ms versus 0.28ms per keystroke on a 241-heading document.

Smaller things: `SETEXT_UNDERLINE` matches every `---` thematic break, so any block above one joins the signature and rebuilds on every keystroke; the ref is mutated during render in `use-editor-outline.ts`; and `mock.module` in `long-document-cost.test.tsx` is never restored, so it leaks to every file loaded after it.

imshashank:
All of it is addressed on `perf/long-documents`, with `main` merged in first. Point by point.

**Stored XSS.** `deferOffscreenImages` is deleted. The sanitiser sets `loading` and `decoding` on `img` inside its element handler, in both the `HTMLRewriter` path and the linkedom path, so the serialiser escapes the values. One deviation from your instruction: I did not add the two names to `ALLOWED_ATTR`. The element handler sets them after `keptAttributes` has filtered, so the allowlist entry is never read for them, and adding it would let `loading` ride on any element (`<p loading="eager">` survived while I had it in). The handler also overrides an author supplied `loading="eager"`, so the value is deterministic.

`packages/services/tests/markdown/lazy-images.test.ts` covers both renderers against both sanitiser paths and asserts on parsed elements and event handler attributes, not substrings. Restoring the splice turns 14 of its 24 cases red, including `eventHandlersIn(html)` returning `["onerror"]` for `<img src="..." alt="a>">`: the injected `loading="` ends the alt value early and the surviving element keeps the handler. Removing the `img` branch from the sanitiser turns 10 red.

**Attributes stripped by the second pass.** Fixed by the same change, and the same test file runs every case through `renderMarkdownWithHeadingIds`, which is the gap that let this through.

**Three untested changes.** All three kept, all three now proven by mutation, none of them mocked:

- `editor.options.extensions` is still the array the editor was created with after twelve keystrokes. Without the `useMemo`, tiptap's `compareOptions` sees a different array every render and calls `setOptions`, which reaches `view.updateState`. That is the reason the memo is worth keeping, and the test goes red without it.
- `editor.options.content` is still the HTML the editor opened with after twelve keystrokes.
- an image mounted in the editor renders with `loading="lazy" decoding="async"`; ProseMirror re-serialises from node attrs, so without `Image.configure` the attributes are gone from the DOM.

**`headingSignature` under-triggering.** Rewritten. Block prefixes (blockquote markers, list markers, indentation) are stripped before a line is tested, raw HTML lines pull in the contiguous block around them, and each chosen line is keyed by its position so a heading freed by a blank line closing an HTML block above it counts as a change. `outline-memo.test.ts` runs your five cases both as insertions and as in-place retitles, plus the blank line case, and asserts the memo agrees with a full render rather than asserting on the signature string. Reverting the prefix stripping reddens three, reverting the raw HTML rule two, dropping the position key one.

**`SETEXT_UNDERLINE`.** A run of dashes only counts as an underline when the line above is an ordinary paragraph, which is the only case CommonMark treats as a setext heading. A break under a list, a heading, a fence or a blank line is left out of the signature entirely, and `> Title` / `> ---` still triggers.

**Ref mutated during render.** The memo is state now, adjusted during render the way React documents. Identity is unchanged, so the scroll spy still does not resubscribe.

**`mock.module`.** Gone from `long-document-cost.test.tsx`. Nothing in that file mocks anything now.

**The numbers.** The eight row table is out. `bun run bench:outline` is committed and reproduces the one claim that survives: 196.52ms per keystroke rebuilt every time versus 0.19ms memoised, on a 241 heading, 34KB document, which lines up with your 180.8ms versus 0.28ms.

`bun run verify` is green: 890 web tests, 227 services tests.

imshashank:
`origin/main` merged in first (now `f020c86`, which landed while I was working, so
the branch is green against the `main` it is about to land on rather than the one
it forked from). One new commit, `928f36a`. `bun run verify` exits 0: 920 web,
212 services, 507 core, 146 shared, 83 db, 42 mcp, 15 realtime-server, 12
realtime-client.

I re-verified each blocker independently against the merged branch rather than
taking the reply above on trust. One of the two was already closed by an earlier
commit and I have left the code alone; the other was still open and is fixed here.

## Stored XSS: closed, re-confirmed, untouched

I wrote a fresh set of twelve break-out payloads rather than reusing the ones in
the test file: bare, `&gt;`, `&#62;` and `&#x3e;` brackets in `alt`; the same in
`title`; unquoted and single quoted attributes; `<script>` and `<svg onload>`
smuggled through `alt`; a markdown `` with a bracket in the alt text; an
author supplied `loading="eager"` alongside a break-out; and a self closing
`/>` variant. Twelve payloads across both renderers and both sanitiser paths is
48 renders, and the parsed output carries **zero** `on*` attributes and zero
`script` elements. Nothing in the sanitiser or in `lazy-images.test.ts` was
changed.

## Blocker already closed: the three untested production changes

Closed by `c2cdc7e`, no code change needed. Reverting all three together turns
the full web suite red, which is the claim that mattered:

```
916 pass, 4 fail
```

- `Image.configure({ HTMLAttributes: { loading, decoding } })` reverted to
  `Image.configure({ allowBase64: false })` reddens *images in the editor > are
  left for the browser to fetch when they scroll into view*.
- dropping `if (frame !== null) return;` from the scroll spy reddens *measures
  once per animation frame however many scroll events arrive* and *schedules
  again once the frame it was waiting for has run*.
- dropping the `useMemo` around `editorExtensions` reddens *keeps handing tiptap
  the same extensions rather than a fresh set per keystroke*.

For completeness I mutated the other three too: unfreezing `openedWith` reddens
*keeps the html the editor opened with*, and removing the `img` branch from
`applyTagRules` reddens 20 of the 24 cases in `lazy-images.test.ts`. All six
production changes are now load bearing.

## Blocker still open: `headingSignature` still under-triggered

`1e0e069` fixed the ATX, blockquote, list indentation and raw HTML classes, and
that work stands. One class was still open: **a setext heading inside a list
item.**

```
# One

- Listed
  ---
```

That is an `h2` titled "Listed" in the rendered document. `plainParagraph` tested
the line above the underline after stripping blockquote markers only, so
`- Listed` was rejected as "starts with a block marker" and the underline never
entered the signature. Retitling it to `- Renamed` left the signature byte
identical, so the sidebar kept showing "Listed". Worse, typing an underline under
the **last** list item in a document adds a heading without shifting any chosen
line's position key, so the outline stayed a heading short while `goTo` and
`activeHeadingId` kept indexing the live DOM by position: the same misalignment
called out in the original review, reached by a different route.

I fuzzed it rather than guessing at the extent: 3889 before/after pairs built
from 22 markdown block shapes across five container nestings, comparing "the
renderer's heading list changed" against "the signature changed". **232 distinct
stale pairs, every one of them a setext heading inside a list item.** After
`928f36a`: **0**.

The fix is two rules:

- test the setext underline and the line above it after the *same* block prefix
  stripping the ATX path already uses, so `- Listed` counts as a paragraph.
- tell a thematic break from a setext underline by how far the dashes are
  indented. Dashes at a column the paragraph above does not reach close the list
  rather than underline it, so `- item` over `---` stays out while `- item` over
  `  ---` goes in. A run of equals signs is never a thematic break, so it always
  counts, which also covers the lazy `- Listed` over `===` that marked renders as
  an `h1`.

That second rule is what keeps the perf guarantee: *does not rebuild the outline
while the block above a break is typed* still passes.

Six new tests, each watched red first and each pinned to a distinct half of the
fix:

| mutation | red |
| --- | --- |
| skip the block prefix stripping in `plainParagraph` | 7 |
| drop the dash indentation rule | 3 |
| restore the signature the branch started with | 17 (was 15 before these tests) |

## Description corrected

The coverage figure was understated. Restoring the pre-fix `outline.ts` reddens
15 tests against the suite as it stood, not 7, and 17 against the suite now. The
description carries the measured numbers and the per-rule breakdown.

## Bench harness confirmed

`bun run bench:outline` runs from a clean checkout of this branch with nothing
but `bun install`. I moved `.env` out of the tree to prove it needs no
environment, no database and no docker, and it still ran. The document shape
reproduces exactly (`34396 bytes, 241 headings`) but the two timings are wall
clock and move run to run, so the description no longer quotes a single figure as
if it were fixed output. Four consecutive runs: 195.04, 202.23, 204.19 and
203.68ms rebuilt against 0.19, 0.18, 0.19 and 0.23ms memoised.

## One observation, pre-existing, not from this branch

`apps/web/tests/features/docs/docs-sidebar.test.tsx` fails with *useToast must be
used inside ToastProvider* when its file is run on its own, and passes in the
full suite only because one of the eleven other test files that `mock.module` the
toast module leaks its stub in first. It behaves the same way on `807487f` and on
`f020c86`, so it is not this branch's doing and I have left it alone, but it is
the same process global `mock.module` hazard that was flagged here and it is worth
its own issue.

PR #113, opened 2026-08-06
Anchor a doc comment to the passage it is about

imshashank:
A doc comment was attached to the whole document. Now it can point at the passage it is
about, the way Linear and Notion both do.

## The anchor

`doc_comment` gains a nullable `anchor` jsonb column: the quoted text, up to 96
characters of context on either side, and the offset the quote was taken from.
`docCommentAnchorSchema` in `@orbit/shared` validates it on the way in, and the same
schema parses it on the way back out in the query cache and in the realtime delta.
The quote is cut at `DOC_ANCHOR_QUOTE_LIMIT` while the anchor is built, so selecting
the whole document produces something the server accepts rather than a comment that
rolls back on submit. Every cut lands on a whole character, so an emoji sitting on the
quote limit or on the edge of the context window cannot leave half of one in the row,
and the schema refuses an anchor that is not whole text.

Locating the passage again is a pure function, `locateDocAnchor`. It takes the
occurrences of the quote nearest the recorded offset, scores each by how much of the
stored context still surrounds it, and breaks ties by distance from that offset. Text
added above the passage moves it and the anchor follows. Text below it changes and the
anchor holds. The quote itself edited away resolves to nothing, and that is exactly
what makes the comment orphaned. A quote that repeats more times than the candidate cap
without a single scrap of its context surviving reports orphaned too, rather than
picking a passage it cannot tell apart.

## In the editor

Selecting text offers **Comment** in the selection bubble, which seeds the composer
with that passage. Each anchored passage carries a ProseMirror inline decoration with
its comment id, so:

- an anchored comment highlights its passage,
- clicking the comment's quote scrolls the document to the passage,
- clicking the passage marks its comment in the thread.

Editor positions and text offsets are mapped through a span table built from the
document, tested both directions over every position in a document.

The recompute is off the keystroke path: it runs once a burst of edits settles, and a
document with no anchored comment is never walked at all. With 30 anchored comments a
pass costs 0.68ms at 20k characters, 2.10ms at 60k and 3.20ms at 200k, so a sixty
character burst costs one pass instead of 41ms, 126ms and 192ms, and the doc header,
share menu, history, nest picker and export menu stop re-rendering per character.

## What does not change

A comment with no anchor renders exactly as it did. A row written before this, or a
delta from an older client with no anchor field, reads back as unanchored. A reply
carries no passage of its own and is refused one: it belongs to the thread it answers,
which shows the passage above it.

In a doc the reader cannot edit there is nothing to scroll, so the quote is shown but
the control to travel to it is disabled.

## Schema

**The schema changed and has to be pushed before this ships.** Either run
`bun run db:push`, or apply the idempotent
`packages/db/catchup/doc-comment-anchor.sql` against the target database.

## Tests

- `packages/shared/tests/utils/doc-anchor.test.ts`: building an anchor and the quote
  limit at its boundary, finding the passage after an edit above it, below it, when the
  quote repeats and when it repeats past the candidate cap, the distance tiebreak, the
  empty quote guard, and detecting the orphan.
- `packages/core/tests/content/doc-comment-service.test.ts`: create and list against
  real Postgres, anchored and unanchored side by side, the reply rule and the message it
  refuses with, the invalid anchor, and the anchor on the sync action.
- `apps/web/tests/features/docs/doc-anchor-editor.test.tsx`: a live `DocEditor` with
  anchored comments, covering the highlight, the focused passage, the click that marks a
  comment, the scroll, the Comment control in the bubble, the text reported back for
  orphan detection, and the walk count over a burst of edits.
- `apps/web/tests/features/docs/doc-surface-anchors.test.tsx`: `DocSurface` over a
  stubbed transport, so the anchor travels from the wire through `docCommentSchema` into
  a highlight, a scroll, a focused comment, an orphan notice, and an inert quote in a
  read only doc.
- `apps/web/tests/features/docs/editor/anchor-positions.test.ts` and
  `anchor-plugin.test.ts`: offset to position round trip, and the decoration actually
  drawn in a live editor.
- `apps/web/tests/features/docs/doc-comments.test.tsx` against the real query hooks,
  plus `doc-anchor-targets.test.ts`, `editor/rich-text-editor.test.tsx` and
  `lib/query/sync.test.ts` for the surface and the cache.

`bun run verify` is green.


imshashank:
The pure layers here are genuinely well tested: 25 mutations across the anchor utils, the validator, the service, `anchor-positions`, `anchor-plugin`, `doc-comments`, `anchorTargetsOf` and `sync.ts`, and 23 were caught by the test whose name matches. The schema change is fine: nullable jsonb, an idempotent catch-up file matching the existing convention, backward compatible. **I have already applied it to production**, so it is safe to land the code whenever this is ready. Authorization is unchanged and both paths still go through `loadReadableDoc`.

What blocks it is that the entire integration layer, the part that delivers every claim under "In the editor", has no test at all.

## Blocking: nine mutations that each kill a headline feature leave all 869 web tests green

`use-doc-anchors.ts` is 146 new lines and nothing in `apps/web/tests` or `apps/web/e2e` imports it. Each of these was applied alone and the full suite stayed green:

1. `ranges.current = decorationsFor(...)` to `[]` (no passage is ever highlighted)
2. drop the `onSelectPassage(commentId)` call (clicking a passage never marks its comment)
3. make `revealPassage` return immediately (clicking a quote never scrolls)
4. delete `editor.registerPlugin(docAnchorPlugin(ranges))` (the plugin is never installed)
5. drop `onAnchorSelected(anchor)` (Comment in the bubble never seeds the composer)
6. drop `onTextChange?.(...)` (orphan detection is dead, every anchor reads as found forever)
7. delete the `onComment` spread in `doc-editor.tsx` (the Comment control never appears in a real doc)
8. delete the anchor props on `<DocComments>` in `doc-surface.tsx`
9. `commenting={{ ...commenting, targets: [] }}`

`anchor-plugin.test.ts` constructs the plugin directly and assigns `ranges.current` by hand, so it never exercises the code that computes ranges, registers the plugin, listens for clicks or scrolls. The feature can ship completely inert.

Also untested: `docCommentSchema.comment.anchor` in `lib/query/schemas.ts`, the path every anchored comment takes on page load. Forcing it to `null` leaves the suite green, because `doc-comments.test.tsx` mocks the whole `use-doc-comments` module.

## Three reachable defects, all reproduced

**A selection over 2000 characters loses the comment.** `buildDocAnchor` slices the quote with no cap while `docCommentAnchorSchema` enforces 2000, so Cmd+A or three paragraphs produces an anchor the server rejects outright. The user sees the passage chip, writes the comment, submits, and the optimistic entry rolls back with a raw Zod message. Clamp the quote client side at `DOC_ANCHOR_QUOTE_LIMIT`.

**The 512-occurrence cap highlights the wrong passage.** `occurrencesOf` stops after 512 hits, so a quote past that point scores only wrong candidates and returns one confidently rather than reporting orphaned. Reproduced: an anchor at offset 1231 resolves to 1022, in the middle of filler. A short quote in a long document hits this. Either fall back to null when the cap is hit and the context score is zero, or search outward from `anchor.start` first.

**In a read-only doc the "Go to the quoted passage" button is enabled and does nothing.** The reader branch renders `<DocComments>` without the anchor props, so `onRevealPassage` is undefined and the orphan notice never appears either. Disable the control or wire the reader path.

## Performance on the editing critical path

`recompute` is bound to the editor `update` event, so every keystroke walks the document, calls `setAnchorText` with the full text, runs `locateDocAnchor` per anchored comment, and dispatches an extra transaction. `setAnchorText` lives in `LoadedDoc`, which previously never re-rendered while typing, so the doc header, share menu, history, nest picker and export menu now re-render per keystroke too. Measured pure anchor work with 30 anchored comments: 1.99ms at 20k characters, 5.02ms at 60k, 7.13ms at 200k, before any React work. Nothing is debounced or memoized.

## Smaller

- The distance tiebreak in `locateDocAnchor` is untested: reversing it changes results and all 16 tests still pass.
- The empty-quote guard is untested.
- The PR says a reply "takes the passage of the comment it answers", but the code stores null and nothing reads the parent's anchor, so a reply shows no passage. Either inherit it or reword the claim and the error message.

imshashank:
All of it is addressed. `main` is merged in and `bun run verify` is green.

## The nine mutations

`apps/web/tests/features/docs/doc-anchor-editor.test.tsx` mounts a real `DocEditor` with a real
`commenting` object and a live tiptap editor, and
`apps/web/tests/features/docs/doc-surface-anchors.test.tsx` renders `DocSurface` over a stubbed
transport, so the anchor travels from the HTTP payload through `docCommentSchema` into the editor.
Each mutation was applied alone against the whole web suite:

| mutation | tests that go red |
| --- | --- |
| `ranges.current = []` | 10 |
| drop `onSelectPassage(commentId)` | 2 |
| `revealPassage` returns immediately | 2 |
| delete `editor.registerPlugin(docAnchorPlugin(ranges))` | 10 |
| drop `onAnchorSelected(anchor)` | 2 |
| drop `onTextChange?.(...)` | 3 |
| delete the `onComment` spread in `doc-editor.tsx` | 2 |
| delete the anchor props on `<DocComments>` | 4 |
| `commenting={{ ...commenting, targets: [] }}` | 4 |

`docCommentSchema.comment.anchor` is covered too: forcing it to `null` now fails 10 tests.
`doc-comments.test.tsx` no longer mocks `use-doc-comments`, it drives the real hooks over a stubbed
`fetch`. That mock was also leaking into every other file in the run, which is how it hid the schema.

## The three defects

**Long selection.** `buildDocAnchor` cuts the quote at `DOC_ANCHOR_QUOTE_LIMIT` and takes the suffix
from the cut, not from the selection that was thrown away. Three tests, including the exact boundary,
so an off by one in the clamp is caught as well.

**The 512 cap.** Candidates are now taken nearest first from `anchor.start`, walking outward in both
directions, so the recorded position is always considered. A search that fills the cap without a
single character of the stored context matching returns `null` instead of picking a passage it cannot
tell apart. The repro is a test: a quote repeated 600 times with the real one behind all of them
resolves to the real one, and reverting to the old scan makes it red.

**Read only doc.** The quote control is disabled unless a surface can actually reveal the passage,
and its label says so instead of promising travel that never happens.

## Performance

The recompute is off the keystroke path. It debounces by 120ms, and a doc with no anchored comment is
not walked at all. Measured with 30 anchored comments: 0.68ms per pass at 20k characters, 2.10ms at
60k, 3.20ms at 200k. A sixty character burst now costs one pass instead of 41ms, 126ms and 192ms, and
a doc with no anchored comment costs nothing per keystroke instead of a walk plus a `LoadedDoc` state
update. Removing the debounce makes the call count test red.

## Smaller

The distance tiebreak and the empty quote guard both have tests now: reversing the tiebreak and
deleting the guard each go red.

On the reply, I took the second option and reworded rather than inheriting. Storing null is the
behaviour, so the refusal now reads "A reply cannot point at a passage of its own." and the test
asserts that message and reads the thread back to show the reply carries none while the root keeps
its own. The PR description says the same thing.

imshashank:
The lone surrogate is fixed and both coverage gaps are closed. `main` is merged in and
`bun run verify` is green. Every test below was watched red before its fix, and red again
under the mutation named next to it.

## The lone surrogate

`buildDocAnchor` sliced at fixed UTF-16 code unit offsets, so a boundary that landed inside a
surrogate pair left half a character in the row. Every cut now opens and closes on a whole
character: a start that would split a pair moves forward, an end that would split one moves
back. Snapping only ever shrinks, so the 2000 and the 96 both still hold, and quote, prefix
and suffix stay exact substrings of the document, which is what `locateDocAnchor` matches on.
`toWellFormed()` would have kept the row insertable but replaced the half character with
U+FFFD, and the anchor would then fail to find a passage that is still right there.

Three boundaries were reachable, not two. The third is the far edge of the context window:

| boundary | mutation that reverts it | tests that go red |
| --- | --- | --- |
| quote clamp at `from + 2000` | drop the snap on `to` | 2 |
| context window opening at `from - 96` | drop the snap on the prefix start | 2 |
| context window closing at `to + 96` | drop the snap on the suffix end | 2 |

Each pair is one unit test in `packages/shared/tests/utils/doc-anchor.test.ts` and one
integration test in `packages/core/tests/content/doc-comment-service.test.ts` that goes through
`createDocComment` against real Postgres, with an emoji sitting exactly on the boundary. Before
the fix the integration tests fail with the error you quoted, `invalid input syntax for type
json`, detail `Unicode low surrogate must follow a high surrogate`.

## The schema, so it is not only the client that is safe

An anchor posted by anything other than this editor could still carry half a character, and
that was the same 500. `docCommentAnchorSchema` now refuses a quote, prefix or suffix that is
not whole text, so a hand written anchor is a 422 with a message rather than a database
failure. Removing the refinement on `quote` makes 2 tests red, removing the one on the context
fields makes 1 red, and a well formed anchor carrying an emoji still parses.

## The DrizzleQueryError hole, kept separate

Worth more than a status code: `toDomainError` put the error's own message in the response, so
a driver failure sent the whole SQL statement and its parameter values to the caller. The
insert that started this shipped the comment body and the anchor JSON back out. `toResponse`
now answers the generic internal error for anything that is not a `DomainError` and logs the
detail server side. A `ZodError` is still 422, a `DomainError` still carries its own message
and status, and the status for a query failure stays 500 because it is a server fault, not
something the caller could have sent differently. Two tests in
`apps/web/tests/lib/api/handler.test.ts`: the old mapping makes the leak test red, and mapping
everything to the generic error makes the domain error test red.

## Coverage gap 1: the zero target fast path

`apps/web/tests/features/docs/doc-anchor-editor.test.tsx` now types a burst into a doc that
carries no anchored comment and asserts the editor reported no text at all, which is the half
of the perf claim that was unprotected. Deleting the fast path in `recompute` makes it red.

## Coverage gap 2: `anchor: input.anchor ?? null`

`apps/web/tests/features/docs/doc-comments.test.tsx` holds the POST open, posts a comment with
a pending passage and reads the quote off the optimistic entry, so the passage is proved to be
on the comment the writer sees before the server answers. Forcing that line to `null` makes it
red.

923 web tests, 0 failures, `bun run verify` exits 0.

PR #158, opened 2026-08-07
docs: add NEXT_PUBLIC_MCP_URL to env example

### LINKED ISSUE #131: NEXT_PUBLIC_MCP_URL is read by the app but missing from .env.example
`apps/web/src/lib/env.ts` reads `NEXT_PUBLIC_MCP_URL`:

```ts
export function mcpServerUrl(): string {
  const explicit = process.env['NEXT_PUBLIC_MCP_URL'];
  if (explicit !== undefined && explicit.trim().length > 0) {
    return mcpUrlSchema.parse(explicit).replace(/\/+$/, '');
  }
  return absoluteUrl('/mcp').replace(/\/+$/, '');
}
```

It is not in `.env.example`, so the only way to discover it is by reading the source. It is documented in [docs/configuration.md](https://github.com/Noveum/orbit/blob/main/docs/configuration.md), but `.env.example` is what people actually copy.

**The fix**

Add it to `.env.example`, commented out with an empty value since the default is correct for almost everyone, near the other `NEXT_PUBLIC_` entries.

**Worth doing at the same time**

Check the rest of `.env.example` against what the code reads. At least one other variable is read directly rather than through the schema: `SLACK_SIGNING_SECRET`, in `apps/web/src/app/api/webhooks/slack/route.ts`. A sweep for `process.env[` across `apps/web/src` and `packages/` will find any others.

Anything you find that is read but not documented belongs in this fix.

Small, useful, and it touches the file every new contributor copies on their first day.

slegarraga:
Closes #131

## Summary

Adds the missing `NEXT_PUBLIC_MCP_URL` to `.env.example`, commented out with an empty value next to the other `NEXT_PUBLIC_` entries. The app already reads it in `apps/web/src/lib/env.ts` and falls back to `<app>/mcp` when unset, so the default remains correct for almost everyone.

## Verification

- Confirmed `apps/web/src/lib/env.ts` reads `NEXT_PUBLIC_MCP_URL`.
- Confirmed the variable is already documented in `docs/configuration.md`.
- Swept `process.env[...]` reads in `apps/web/src` and `packages`; the other main runtime variables in the sweep are already present in `.env.example`.

Docs-only change.

Signed off with DCO.

<h3>Greptile Summary</h3>

Adds the previously missing `NEXT_PUBLIC_MCP_URL` entry to the environment template without changing the application's default behavior.

- Documents that the variable overrides the advertised MCP endpoint.
- Leaves the optional variable commented out and defaults to `<app>/mcp`.

The PR appears safe to merge with no actionable issues identified.

The new environment-template entry accurately reflects existing behavior, and leaving it commented or empty preserves the established MCP endpoint fallback.

| Filename | Overview |
|----------|----------|
| .env.example | Adds an accurate commented example for the optional MCP URL override; no runtime or configuration issue identified. |

</details>

<sub>Reviews (1): Last reviewed commit: ["docs: add NEXT\_PUBLIC\_MCP\_URL to env exa..."](https://github.com/noveum/orbit/commit/f102cccb4aae23cf1b6d95320b1365f71cdb92f5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51186319)</sub>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,8 @@ NEXT_PUBLIC_APP_URL=http://localhost:3000
 
 NEXT_PUBLIC_REALTIME_URL=ws://localhost:3100
 REALTIME_PORT=3100
+# Overrides the advertised MCP endpoint. Defaults to <app>/mcp.
+# NEXT_PUBLIC_MCP_URL=
 
 GOOGLE_CLIENT_ID=
 GOOGLE_CLIENT_SECRET=


slegarraga:
Verification for this docs-only change: `bun run check-comments` and `bun run check-bytes` pass, `bun run typecheck` passes, and `bun run lint` reports only the existing warning in `packages/db/tests/check-source-bytes.test.ts`. The full test suite needs local Postgres/Redis, so I did not run it for this env-file change.

imshashank:
Thanks for this, and sorry for the duplicate: #146 landed first and covers the same `NEXT_PUBLIC_MCP_URL` documentation, plus `DATABASE_POOL_MAX` which was also undocumented.

Both variables were real and neither was in `.env.example`, so the gap you spotted was genuine. It is fixed on main now.

Closing this as already covered rather than because anything was wrong with it.

PR #161, opened 2026-08-07
fix: wire clear filters callback on My Issues view

### LINKED ISSUE #152: Clear filters does nothing on My issues
`MyIssuesView` renders `HiddenFooter` with a clear-filters callback that does nothing:

```ts
// apps/web/src/features/issues/my-issues-view.tsx
<HiddenFooter
  hiddenByFilters={model.hiddenByFilters}
  hiddenByDisplay={model.hiddenByDisplay}
  onClearFilters={() => undefined}
  ...
/>
```

`HiddenFooter` shows a **Clear filters** action whenever `hiddenByFilters > 0`. On My issues that action is inert: clicking it leaves the filter in place and the hidden issues hidden.

**How to see it**

A filter can reach the page through the URL, since `useViewConfig` reads its state from the query string. With one applied and some issues hidden, the footer offers to clear it and then does not.

**The fix**

Clear the filter, the way the standup board now does after the same bug was caught there in #149:

```ts
onClearFilters={() => setConfig({ ...config, filter: { ...config.filter, children: [] } })}
```

**Worth checking while you are in there**

Whether any other surface passes a no-op to `onClearFilters`. `grep -rn "onClearFilters" apps/web/src` finds them all. A footer that offers an action it does not perform is worse than no footer.

Found while fixing the copy of this bug in the standup board, so the correct behaviour is already in the tree to copy from.

lb1192176991-lab:
## What

Wired the onClearFilters callback in MyIssuesView to actually clear filters, matching the behavior already implemented in StandupBoard and SavedViewPage.

## Why

The HiddenFooter component shows a Clear filters action when hiddenByFilters > 0, but on My Issues the callback was () => undefined, making the button inert. Users could click it but nothing happened.

## Testing

Changed onClearFilters={() => undefined} to onClearFilters={() => setConfig({ ...config, filter: { ...config.filter, children: [] } })}, following the exact same pattern used by the standup board and saved views.

Closes #152

<h3>Greptile Summary</h3>

The PR makes the “Clear filters” action functional in the My Issues hidden-items footer.
- Replaces the no-op callback with a configuration update that empties the filter group.
- Preserves the filter group’s existing kind and combinator.
- Matches the established clear-filter behavior used by sibling issue views.

The PR appears safe to merge, with no actionable regressions identified.

The callback now performs the expected clear-filter operation using the same configuration transformation as sibling views, while preserving unrelated filter and display settings.

| Filename | Overview |
|----------|----------|
| apps/web/src/features/issues/my-issues-view.tsx | Correctly wires the footer action to clear active filter conditions without changing unrelated view configuration. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: wire clear filters callback on My I..."](https://github.com/noveum/orbit/commit/3d49658f4fa7b9856a1eaf29f9b47d18a86f1f6e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51265863)</sub>

### DIFF
--- a/apps/web/src/features/issues/my-issues-view.tsx
+++ b/apps/web/src/features/issues/my-issues-view.tsx
@@ -127,7 +127,7 @@ export function MyIssuesView() {
       <HiddenFooter
         hiddenByFilters={model.hiddenByFilters}
         hiddenByDisplay={model.hiddenByDisplay}
-        onClearFilters={() => undefined}
+        onClearFilters={() => setConfig({ ...config, filter: { ...config.filter, children: [] } })}
         onRevealDisplay={() =>
           setConfig({
             ...config,


imshashank:
Good catch, and the fix is right. `onClearFilters` was `() => undefined` on My Issues while `standup-board.tsx:119` and `saved-view-page.tsx:149` both do exactly what you wrote, and `team-view.tsx:141` does it through a named helper. The button rendered and did nothing.

Merging on that basis. One thing I want to be honest about rather than quietly skip: this repo's rule is that a feature is not done until it has a test that would fail if the feature broke, and I could not get a meaningful one to land here. Filtering on this view is server side, so the query key changes and the cached fixture no longer applies; asserting through the URL sync does not work either, because `use-view-config` skips the write when `window.location.pathname` does not match the mocked `usePathname`, which is the case under happy-dom. I reverted my attempts rather than leave a test that passes for the wrong reason.

Tracked separately so it does not get lost. Thanks for the fix.

PR #263, opened 2026-08-09
feat(webhooks): scope a delivery to the workspace it belongs to

imshashank:
## Why

#260 surfaced why a GitHub delivery changed nothing, but `webhook_delivery` had no workspace column, so the list could only be gated on `integration:manage` and still showed **every** workspace's traffic to whoever could read it. On a deployment with many workspaces that is both noisy and wrong.

## The design decision

The obvious move is to attribute a delivery by its repository, since that is what `applyGithubEvent` already resolves. That is exactly backwards for diagnostics: `repository_not_connected` is the single most useful reason to show a workspace, and repository-based attribution **cannot produce it by definition**, because there is no repository row to attribute through.

So attribution goes through the **installation**. Every GitHub App payload carries `installation.id`, and `github_installation` maps it to a workspace.

| Case | Attributed | Visible to the workspace |
| --- | --- | --- |
| Repository connected here | yes | yes |
| Repository not connected, installation is ours | **yes** | **yes**, with the reason |
| Installation belongs to another workspace | no | no |
| Payload with no installation, or unparseable | no | no |

## Does this work for everybody, not just this deployment

This is the part I went back and checked properly, because Orbit is installed by other people in other places.

**Both documented install paths apply the column.** The repository now has two, and the docs describe only one:

- `bun run db:push`, which is what `README.md`, `CONTRIBUTING.md`, `docs/getting-started.md` and `docs/self-hosting.md` all tell people to run. Push applies the schema directly, so the column lands.
- `bun run db:migrate`, which is what CI exercises since #256 and what a fresh clone needs. This PR adds `0003_productive_quicksilver.sql` so that path lands too.

I verified the second the way CI does rather than assuming: built a database from **nothing but the committed migrations** and ran the drift check against it.

```
$ createdb orbit_from_migrations && bun run --filter '@orbit/db' migrate
[OK] migrations applied successfully!
$ bun run db:check-drift
localhost/orbit_from_migrations has every table and column the schema declares.
```

**The migration is additive and safe on an existing install.** A nullable `text` column, its foreign key, and an index. No table rewrite, no backfill, no default to compute. On Postgres 11 and later `ADD COLUMN` without a volatile default is O(1). The foreign key validates existing rows, but every existing row is null, so there is nothing to check. Retention already bounds the table: #253 added a 30 day prune after it reached 23,000 rows.

**Old rows degrade in the safe direction.** Deliveries recorded before this carry no workspace and stop appearing, rather than appearing under the wrong one.

**A deployment with no `github_installation` rows loses nothing.** Its deliveries are unattributed and the list is empty, but that install is already told plainly what is wrong: the integrations page says GitHub is not connected. No diagnostic disappears, it moves somewhere better.

## Tests

2 new against a real database: a delivery for an unconnected repository is stamped with the workspace behind its installation and still records `repository_not_connected`, and a delivery whose installation belongs elsewhere stays unattributed. 28 pass on that route.

`bun run verify` green.

## One thing worth deciding separately

Every install doc still says `bun run db:push`, including against a production database. Since #256 the repository also carries committed migrations and CI refuses a schema they do not build. Both work today, so nothing is broken, but a self-hoster following the docs will never run `db:migrate`. That belongs with whoever owns #256 rather than in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

The PR attributes GitHub webhook deliveries to workspaces through GitHub App installations and displays workspace-scoped delivery diagnostics.
- Adds nullable workspace attribution to webhook deliveries, including migrations and an index.
- Resolves and records the workspace before processing signed GitHub webhook payloads.
- Loads the latest workspace-scoped deliveries for integration administrators.
- Adds delivery-list UI and database-backed route/component tests.

<h3>Confidence Score: 4/5</h3>

The PR is not yet safe to merge because the delivery summary can report that every recent webhook changed Orbit even when one failed or completed without a state change.

The existing summary branch still treats every non-ignored delivery as a successful mutation, while current webhook processing can persist failed deliveries and processed deliveries with no synchronization actions.

**Files Needing Attention:** apps/web/src/features/settings/github-deliveries.tsx

| Filename | Overview |
|----------|----------|
| apps/web/src/app/api/webhooks/github/route.ts | Resolves the signed payload's installation and stamps the corresponding workspace onto the delivery before event handling. |
| apps/web/src/features/settings/integrations-data.ts | Adds permission-gated, workspace-scoped loading of the 20 latest GitHub deliveries. |
| apps/web/src/features/settings/github-deliveries.tsx | Adds the recent-deliveries presentation and human-readable reason labels. |
| packages/db/src/schema/comms.ts | Adds nullable workspace attribution and its lookup index to the webhook-delivery schema. |
| packages/db/drizzle/0004_clear_mandroid.sql | Migrates persisted webhook deliveries with the nullable workspace foreign key and composite index. |

</details>

```mermaid
sequenceDiagram
    participant GitHub
    participant Webhook as GitHub webhook route
    participant Installations as github_installation
    participant Deliveries as webhook_delivery
    participant Settings as Integration settings
    GitHub->>Webhook: Signed event with installation.id
    Webhook->>Installations: Resolve installation globally
    Installations-->>Webhook: organizationId or null
    Webhook->>Deliveries: Stamp organizationId
    Webhook->>Webhook: Process event and record status
    Settings->>Deliveries: Query latest rows by organizationId
    Deliveries-->>Settings: Workspace-scoped deliveries
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Merge main, and take the migration slot ..."](https://github.com/noveum/orbit/commit/fcd7792fec82d5e72f6de2fa09f8bbd38d7ee847) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51546555)</sub>


imshashank:
@coderabbitai review

PR #274, opened 2026-08-09
Give people a handle they can be mentioned by

imshashank:
Reopens the fix from #267, which was closed on the understanding that `handleFor` used a user ID rather than a random string. It doesn't. There are two functions with that name, and the one that runs at sign-up still appends ten random characters unconditionally on `main` today:

```ts
// apps/web/src/lib/auth/server.ts
function handleFor(email: string, name: string): string {
  const base = slugify(name) || slugify(email.split('@')[0] ?? '') || 'member';
  return `${base}-${randomUUID().replaceAll('-', '').slice(0, 10)}`;
}
```

The other one, `packages/db/src/import/plane-mapping.ts`, takes a `taken` set and already does clean base-plus-numbering. That is the good one, and it is almost certainly the one that was looked at. It only runs on import, which is why imported accounts have usable handles and everyone who signed up does not.

Handles are how mentions resolve, so in practice nobody who signed up could be @-mentioned. An account invited as `hey@yodu.ai` came out as `hey-3402a5fa15`, which no colleague is going to type.

## What changes

The base handle stands on its own. A real collision takes `ada-lovelace-2`, then `-3`, and the random suffix survives only as a last resort once twenty numbered candidates are gone. All candidates are checked in one query.

Generation moves to `apps/web/src/lib/auth/handle.ts` so it can be tested without standing up auth: the caller passes in the "which of these are taken" lookup, and the sign-up hook supplies the database one.

Existing accounts are not touched. Anyone already carrying a random suffix can change it in profile settings, which already validates uniqueness.

## The race, and why it is not fixed here

The previous review flagged that the availability query and the insert are separate operations, so two simultaneous sign-ups sharing a base can both select the same free handle and one insert then fails on the unique constraint. That is correct, and it is inherent to a Better Auth `user.create.before` hook: the hook returns the row it wants inserted, so there is no insert error to catch and retry from.

I have deliberately not restructured the hook to fix it, because the trade is lopsided. Today **every** sign-up gets an unusable handle, with certainty. After this, a sign-up can fail only when two people with the same slugified name submit within the same instant, and that failure is retryable. Trading a universal defect for a rare retryable one is worth doing on its own; fixing the race properly means assigning the handle after insert and living with a window where it is wrong, which is a bigger change than this bug warrants.

Worth its own issue if it ever actually bites.

## Testing

Nine cases in `apps/web/tests/lib/auth/handle.test.ts` covering the name and email bases, the `member` fallback, first-come clean handles, numbering on collision, and the random-suffix last resort.

Verified by mutation rather than asserted: restoring the unconditional suffix fails **4 of the 9**, with the file hash checked before and after to prove the mutation actually applied and was reverted.

`bun run typecheck` passes. `bun run lint` is clean. The two failures in `tests/lib/auth/sign-out.test.ts` are realtime socket tests that need a live database and fail identically on `main`, so they are not from this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

This PR replaces unconditional random handle suffixes with a batched availability lookup that prefers clean handles and then numbered alternatives. The previously reported minimum-length mismatch remains in the current implementation.
- Extracts handle generation into a separately testable module.
- Queries all clean and numbered candidates in one database operation.
- Adds tests for base selection, collisions, numbering, and random fallback.

<h3>Confidence Score: 4/5</h3>

The PR should not merge until generated handles satisfy the mention parser’s minimum length.

A signup whose name or email local part slugifies to one character still persists that value as the handle, while mention extraction only recognizes handles containing at least two characters.

**Files Needing Attention:** apps/web/src/lib/auth/handle.ts

| Filename | Overview |
|----------|----------|
| apps/web/src/lib/auth/handle.ts | Introduces clean and numbered handle generation, but still permits one-character handles that cannot be resolved by mention extraction. |
| apps/web/src/lib/auth/server.ts | Connects asynchronous uniqueness lookup to the user creation hook and persists the selected handle directly. |
| apps/web/tests/lib/auth/handle.test.ts | Covers generation and collision behavior but does not cover the outstanding one-character handle case. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(auth): name the handle-candidate con..."](https://github.com/noveum/orbit/commit/62d5ca9172590c4ae579b3586eb1692aa4b967bf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51625841)</sub>

### DIFF
--- a/apps/web/src/lib/auth/handle.ts
+++ b/apps/web/src/lib/auth/handle.ts
@@ -1,0 +1,28 @@
+import { randomUUID } from 'node:crypto';
+import { slugify } from '@orbit/shared/utils';
+
+const NUMBERED_CANDIDATES = 20;
+
+export function handleBaseFor(email: string, name: string): string {
+  return slugify(name) || slugify(email.split('@')[0] ?? '') || 'member';
+}
+
+function candidatesFor(base: string): string[] {
+  const numbered = Array.from(
+    { length: NUMBERED_CANDIDATES - 1 },
+    (_unused, index) => `${base}-${index + 2}`,
+  );
+  return [base, ...numbered];
+}
+
+export async function uniqueHandleFor(
+  email: string,
+  name: string,
+  taken: (candidates: readonly string[]) => Promise<Set<string>>,
+): Promise<string> {
+  const base = handleBaseFor(email, name);
+  const candidates = candidatesFor(base);
+  const used = await taken(candidates);
+  const free = candidates.find((candidate) => !used.has(candidate));
+  return free ?? `${base}-${randomUUID().replaceAll('-', '').slice(0, 10)}`;
+}
--- a/apps/web/src/lib/auth/handle.ts
+++ b/apps/web/src/lib/auth/handle.ts
@@ -1,15 +1,15 @@
 import { randomUUID } from 'node:crypto';
 import { slugify } from '@orbit/shared/utils';
 
-const NUMBERED_CANDIDATES = 20;
+export const HANDLE_CANDIDATES = 20;
 
 export function handleBaseFor(email: string, name: string): string {
   return slugify(name) || slugify(email.split('@')[0] ?? '') || 'member';
 }
 
-function candidatesFor(base: string): string[] {
+export function candidatesFor(base: string): string[] {
   const numbered = Array.from(
-    { length: NUMBERED_CANDIDATES - 1 },
+    { length: HANDLE_CANDIDATES - 1 },
     (_unused, index) => `${base}-${index + 2}`,
   );
   return [base, ...numbered];
--- a/apps/web/src/lib/auth/server.ts
+++ b/apps/web/src/lib/auth/server.ts
@@ -1,4 +1,3 @@
-import { randomUUID } from 'node:crypto';
 import { passkey } from '@better-auth/passkey';
 import {
   assertEmailDomainAllowed,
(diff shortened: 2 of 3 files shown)


imshashank:
Both findings are valid. One is fixed, one is deliberately not.

**Candidate count (minor).** Fixed in `62d5ca9`, though as naming rather than behaviour. The count was already right — base plus `-2` through `-20` is twenty candidates — but the constant was called `NUMBERED_CANDIDATES`, which claims to count only the numbered ones. Renamed to `HANDLE_CANDIDATES` and pinned with two tests: the list is exactly `HANDLE_CANDIDATES` entries starting at the bare base, and the random suffix appears only once every one of them is taken. Adding a 21st candidate would have made the code match a misleading name instead of fixing it.

**Atomic allocation (major).** Correct, and not fixed here on purpose.

It cannot be fixed inside the current hook. `handleFor` runs in Better Auth's `user.create.before`, which returns the row it wants inserted — Better Auth performs the insert, so there is no insert error in scope to catch and retry from. Making it atomic means moving allocation after the insert and reconciling, which leaves a window where the handle is wrong and is a larger change than this bug justifies.

The trade is lopsided in favour of shipping:

- today, **every** sign-up gets an unusable handle, with certainty
- after this, a sign-up can fail only when two people whose names slugify identically submit within the same instant, and that failure is retryable

Trading a universal defect for a rare retryable one is worth doing on its own. I would rather leave the race documented than hold a fix for a certain bug behind a restructure aimed at an uncommon one.

Worth its own issue if it ever actually bites.

Issue #194, opened 2026-08-08
The audit log table is never written to


imshashank:
`audit_log` exists in `packages/db/src/schema/comms.ts` with a well designed set of columns: actor type, actor id, actor name, action, entity type, entity id, a `before` and `after` JSONB pair, and an IP address, indexed on organisation and time.

Nothing writes to it. Not one row, ever.

This matters more than the other dormant tables because of what people assume when they see it. An admin evaluating Orbit for a team, reading the schema, will conclude there is an audit trail. There is not. If somebody removes a member, changes a role, deletes a project or rotates an integration token, there is no record of it anywhere.

`issue_activity` covers changes to issues and is a different thing. It is a product feature for showing history on an issue, scoped to issues, and it does not record administrative actions at all.

## What to build

**Write the log.** Every administrative and destructive action produces a row:

- Member invited, role changed, removed
- Team created, renamed, deleted, membership changed
- Workspace settings changed, including the allowed email domains
- Integration connected, disconnected, token rotated
- Project or doc deleted, archived, or its visibility changed
- Doc share link created or revoked
- API key created or revoked, once those exist
- Sign-in method changes, and session revocation

The `before` and `after` columns mean the log can answer what a setting used to be, which is usually the actual question. Use them. Never put a secret, a token or a password hash in either one.

**Read the log.** An admin only view under settings, filterable by actor, action, entity and date, with the newest first. The index `audit_log_org_idx` on organisation and time already supports that ordering.

**Retain the log.** Decide and document a retention policy. Unbounded growth on a busy workspace is a real problem, and a self hoster needs to know whether this table will grow forever. Whatever you pick, put it in [docs/self-hosting.md](https://github.com/Noveum/orbit/blob/main/docs/self-hosting.md).

## Things to get right

- **The log is append only.** No update path, no delete path, no API that can rewrite history. If retention deletes old rows, that is a scheduled job and not a user action.
- **`actor_type` defaults to `user` and the column exists for a reason.** MCP agents and automation rules act too, and an audit entry that attributes an agent's action to the human who authorised its token is misleading. Record both: the agent as actor, the granting user as the principal behind it. The `mcp_grant` row already binds a client and user to a workspace, so the information is available.
- **The IP address column is personal data.** Say in the docs that it is collected, and consider making it optional for self hosters who would rather not.
[...]

Issue #191, opened 2026-08-08
Send personal notifications as a Slack DM, not to a channel


imshashank:
Orbit's Slack notifications go to channels. `slack_channel_sync` in `packages/db/src/schema/comms.ts` routes a team's activity to that team's channel, which is right for team-wide events and wrong for anything addressed to one person.

"You were assigned SEC-42" and "You were mentioned in a comment" are personal notifications. Delivering them to a channel means everybody reads what one person needed to know, and the person who needed to know it has to find it in a stream.

## What to build

Deliver personal notifications as a Slack DM from the Orbit app to the mapped user.

The routing rule, roughly:

- Notifications with a single intended recipient go to a DM. Assignment, mention, comment on something you are subscribed to, a review request.
- Notifications about a team or project as a whole go to the channel. Sprint started, project status changed, a release cut.
- Where both make sense, the person's preferences decide.

## Preferences

`notification_preference` and `notification_setting` already exist in `packages/db/src/schema/comms.ts`, and the matrix UI is at `apps/web/src/features/settings/notification-matrix.tsx`. It already has per event and per channel controls, and quiet hours already work across every channel.

So the work is mostly:

1. Add a Slack DM destination alongside the existing channels, in the preference model and in the matrix.
2. Route to it in the notification producers under `packages/core/src/notifications/`.
3. Respect quiet hours, which the existing code already does for other channels, so follow it rather than reimplementing.

## Requirements

- **`im:write` scope.** This is a new bot scope and it has to be added to the app and to the setup instructions in [docs/integrations.md](https://github.com/Noveum/orbit/blob/main/docs/integrations.md). Anyone whose workspace installed the app before this lands will need to reauthorise, so the settings page has to detect a missing scope and say so rather than failing silently. That reauthorisation path is part of this issue.
- **Unmapped users get nothing and it is not an error.** If a person has no Slack account matched, the DM channel is simply unavailable to them and the UI should show it as unavailable rather than as off.
- **Do not double deliver.** Somebody with the inbox, email and Slack DM all enabled for assignment should get three notifications only if they asked for three. Check what the current behaviour is across channels before changing it.
- **Never DM about something the person may not read.** The audience calculation in `packages/core/src/notifications/audience.ts` is what decides who is told, and it stays the authority.

## Files

- `packages/core/src/notifications/notify.ts`
- `packages/core/src/notifications/audience.ts`
- `packages/db/src/schema/comms.ts`
[...]

Issue #183, opened 2026-08-08
Cycle time and lead time never say what they measure


imshashank:
On a sprint that has not completed anything, the Cycle time and Lead time cards both say "Nothing completed yet" and stop. On a sprint that has completed something, they show a histogram with no explanation of what is being measured, from when to when.

Neither state tells a reader what these numbers mean, and both of them are terms that different teams define differently.

## What the code actually measures

`cycleFlowMetrics` in `packages/core/src/analytics/burndown.ts` is the source of truth, and the UI should say out loud what it does:

- Which timestamps each metric runs between. The issue table has `created_at`, `started_at`, `state_entered_at`, `completed_at` and `canceled_at`, so there are several plausible definitions and only one implemented.
- Whether cancelled issues are counted.
- Whether issues that moved into the sprint late are counted from when they were created or from when they entered the sprint.

Read the function, then write down what it does. If what it does is not what most teams would expect, say so on this thread rather than quietly changing it, because somebody is already reading these numbers.

## What to build

**A better empty state.** "Nothing completed yet" is accurate and useless. Replace it with something that says what would fill it and when: which sprint it is looking at, that cycle time is measured over issues completed inside the sprint window, and that the card fills in once one is done. If the sprint has not started, say that instead.

**A definition on the card.** A short line under the title, or a hover, that states the measurement in one sentence. "Time from first moving into a started state to completion, for the 12 issues completed in this sprint." The count matters as much as the distribution, because a median over three issues is not a median.

**Show the sample size and the spread.** `distributionOf` in `packages/core/src/analytics/math.ts` already computes percentiles. Surface p50 and p90 as numbers next to the histogram rather than leaving the reader to estimate them off the bars. `percentile` is already there and already tested.

**Document it.** Add the definitions to `docs/concepts.md`, which is where Orbit explains what it calls things. Right now cycle time and lead time appear in the product and nowhere in the documentation.

## Files

- `apps/web/src/features/analytics/cycle-panel.tsx`
- `packages/core/src/analytics/burndown.ts`
- `packages/core/src/analytics/math.ts`
- `apps/web/src/features/charts/histogram.tsx`
- `docs/concepts.md`

## Good first issue

The empty state and the definition line are small, self contained and need no database work. The percentile surfacing is slightly more involved. Any of the three is a complete contribution on its own.

imshashank:
Most of this landed in #309. One piece is left, so re-scoping rather than closing.

[...]

Issue #175, opened 2026-08-08
The scope chart renders blank when the series has a single bucket


imshashank:
On a workspace where every issue was created in the same week, the "Scope and completed over time" card draws nothing. The card, the legend and the two date labels all render, and the plot area is empty. It reads as a chart that failed to load, and people report it as one.

It is not a loading failure. It is a one point series being asked to draw a line.

## Why it happens

`scopeSeries` in `packages/core/src/analytics/distribution.ts` buckets issues by week between the first `created_at` and the last completion. A workspace whose history fits inside one week gets exactly one bucket, so `series.length === 1`.

That array reaches `linePath` in `apps/web/src/features/charts/geometry.ts`:

```ts
export function linePath(values: readonly number[], max: number): string {
  if (values.length === 0) return '';
  return values
    .map((value, index) => {
      const command = index === 0 ? 'M' : 'L';
      return `${command}${chartX(index, values.length).toFixed(2)} ${chartY(value, max).toFixed(2)}`;
    })
    .join(' ');
}
```

With one value the output is a single `M160.00 6.00`. A path that is only a move command has no length, so SVG paints nothing. `areaPath` then closes a zero width shape, which also paints nothing.

`ScopeSection` guards on `series.length === 0` and shows a sensible empty state, so the zero case is handled. The one case falls straight through.

## What to do

Draw something for a single point. The two reasonable options, in order of preference:

1. **Render a dot.** When `values.length === 1`, emit a small circle at `chartX(0, 1)` and `chartY(value, max)` instead of a path. This is honest: there is one observation and here is where it sits.
2. **Emit a flat segment.** Repeat the point so the path spans the width. Simpler, but it implies a trend across time that the data does not contain, so prefer the dot.

Whichever you pick, `Histogram` and `BurndownChart` in the same directory share this geometry and have the same hole. Check them and fix them in the same change.

## Files

- `apps/web/src/features/charts/geometry.ts`
- `apps/web/src/features/charts/line-chart.tsx`
- `apps/web/src/features/charts/burndown-chart.tsx` is in `apps/web/src/features/analytics/`
- `apps/web/src/features/charts/histogram.tsx`

## Reproducing it

`bun run db:seed` loads a demo workspace whose issues span enough time to hide this. The quickest reproduction is a unit test:

```ts
expect(linePath([5], 10)).not.toBe('M160.00 6.00');
```

## Tests

[...]

PR #319, opened 2026-08-15
fix(core): protect docs when deleting collections

### LINKED ISSUE #245: Deleting a doc collection relocates pages the deleter cannot write
Found during review of #241 by CodeRabbit. Pre-existing, not introduced there.

## What happens

`deleteDocCollection` checks the workspace level `doc:write` capability, then `emptyCollection` updates every document whose `collection_id` matches, with no per document authorization:

```
packages/core/src/content/doc-service.ts  emptyCollection / deleteDocCollection
```

Elsewhere a write goes through `assertDocWritable`, which honours `doc_access` and the document's visibility. This path does not. So a member who may manage documents can relocate a document they are not allowed to edit by deleting the collection it sits in.

## What it does and does not do

- It changes where a document is filed. Nothing is deleted, no content is read, and no visibility changes.
- It predates #241. `emptyCollection` already moved every document in the collection to unfiled; #241 only added the option to move them to a named collection instead.

## The decision to make

The fix is not just adding a check, it is choosing what should happen when a collection holds a document the deleter cannot write:

1. refuse the whole delete and name the documents in the way, or
2. delete the collection, relocate only the documents the caller may write, and leave the rest where they are, which leaves rows pointing at a collection that no longer exists unless the column is nulled anyway, or
3. restrict collection deletion to a role that can modify every document in the organisation.

Option 1 is the most predictable and the easiest to explain, but it makes a delete fail for a reason the person cannot see or fix. Worth deciding deliberately rather than picking one inside an outage fix.

Whichever is chosen needs a test that a member without write access to a document cannot move it by deleting its collection.

mikemikimike:
## What this changes

Preflight every document in a collection with `assertDocWritable` before deleting the collection and reassigning its documents. If any document is not writable by the caller, the whole operation is rejected before any rows are changed.

## Why

A member with collection-level `doc:write` could previously delete a collection and move documents they were not allowed to edit. Closes #245

## How you know it works

Added a regression test proving a member cannot delete a collection containing a private document owned by another user, and that both the collection and document placement remain unchanged.

- `bun --version`: 1.3.14
- `bun run --filter '@orbit/core' typecheck`: passed
- `bunx biome lint packages/core/src/content/doc-service.ts packages/core/tests/content/doc-service.test.ts`: passed
- `git diff --check`: passed
- The focused database test was attempted after starting Docker and running `bun run db:push` and `bun run db:test-setup`. The test database lane hung while cloning/resetting and timed out, so no passing test result is claimed.

## Checklist

- [ ] `bun run verify` is green, all four checks

The chosen behavior is the issue's option 1: fail the entire collection deletion when any contained document is not writable. The preflight runs in the existing transaction, so no partial reassignment is committed.

### DIFF
--- a/packages/core/src/content/doc-service.ts
+++ b/packages/core/src/content/doc-service.ts
@@ -1551,6 +1551,18 @@ export async function deleteDocCollection(
         .limit(1);
       requireRow(target, 'That collection does not exist.');
     }
+    const docs = await tx
+      .select(DOC_COLUMNS)
+      .from(schema.doc)
+      .where(
+        and(
+          eq(schema.doc.organizationId, principal.organizationId),
+          eq(schema.doc.collectionId, collectionId),
+        ),
+      );
+    for (const doc of docs) {
+      await assertDocWritable(tx, principal, doc);
+    }
     const syncId = await nextSyncId(tx);
     const actor = await principalActor(tx, principal);
     const homeless = await emptyCollection(
--- a/packages/core/tests/content/doc-service.test.ts
+++ b/packages/core/tests/content/doc-service.test.ts
@@ -261,6 +261,22 @@ describe('collections', () => {
       code: 'forbidden',
     });
   });
+
+  it('refuses to move a doc the caller cannot write', async () => {
+    const { collection } = await createDocCollection(workspace.admin, { name: 'Private' });
+    const { doc } = await createDoc(workspace.admin, {
+      title: 'Restricted',
+      collectionId: collection.id,
+      visibility: 'private',
+    });
+    const member = await addMember(workspace, 'member', { name: 'Mina Member' });
+
+    await expect(deleteDocCollection(member.principal, collection.id)).rejects.toMatchObject({
+      code: 'forbidden',
+    });
+    expect(await listDocCollections(workspace.admin)).toHaveLength(1);
+    expect((await getDoc(workspace.admin, doc.id)).doc.collectionId).toBe(collection.id);
+  });
 });
 
 describe('published doc urls', () => {


imshashank (changes requested):
Thanks for this, and for being straight in the description about not having got the test to run. I checked it out and ran it, and I want to lead with what you got right, because the central decision here is a good one.

## Verified working

You route the decision through `assertDocWritable` rather than re-deriving the rules, which is what `packages/shared/src/policy` exists for. That matters more than it might look. I probed every visibility case against a live database:

```
admin deleting a collection holding a member's private doc  -> deletes, correct
a published (public) doc authored by another member         -> deletes, correct
a link shared doc authored by another member                -> deletes, correct
a team doc authored by another member                       -> refused, correct
a private doc authored by another member                    -> refused, correct
a nested private child of another member                    -> refused, correct
an ordinary workspace doc by another member                 -> deletes, correct
```

Every one of those is right, including the ones that are easy to get wrong: the admin bypass survives, and `public` and `link` documents correctly do not block, because `isRestricted` is `private || team`. Reusing the existing predicate is what buys you that.

Also confirmed:

- Your test genuinely fails without the change. I reverted the source file and it fails as expected.
- Asserting that the collection survives and the document placement is unchanged is a good instinct. It proves the transaction actually rolled back, which a bare `rejects` assertion would not.
- The preflight covers exactly the rows `emptyCollection` later updates, nested descendants included, because they inherit `collection_id`.
- `DOC_COLUMNS` rather than a bare `select()`, correct.
- Full suites pass locally: 929 in `@orbit/core`, 193 in `@orbit/mcp-server`. Comment policy, biome and whitespace are all clean. CI has since gone green on unit and integration too, so you can tick that checklist box.

## Please fix before merge

**1. The refusal needs to tell the user something they can act on.**

Right now the blocking document is completely invisible to the person deleting. I confirmed that `listDocs` returns an empty list for them, and then the delete fails with "You only have read access to that doc." So they see an empty folder they cannot delete, for a reason they can neither see nor fix. Any member can also drop a private document into any collection and permanently block its deletion for everyone except an admin.

Issue #245 asked for option 1 as "refuse the whole delete **and name the documents in the way**", and the naming half is what makes it survivable. Please gather the blocking documents rather than throwing on the first one, then raise an error that names the ones the caller can actually read and counts the ones they cannot, something like "2 pages in this folder are not yours to move: Runbook, Protocol, and 1 more you cannot see."

There is a consistency point worth knowing here. `loadReadableDoc` deliberately throws `notFound` rather than `forbidden` for a document the caller cannot read, specifically so the response does not confirm that hidden content exists. This path throws `forbidden` for exactly that case. Counting the invisible ones rather than naming them keeps that intent intact while still being useful.

**2. The race Greptile flagged is real. I reproduced it.**

I replayed your statement sequence against Postgres with two connections:

```
preflight saw 0 docs, so the authorization loop passes trivially
a private doc was committed into the collection by another transaction
the delete path reassigned 1 docs
the unchecked private doc was reassigned: true
```

So a document committed into the collection after the preflight gets reassigned having never been authorized, which is the original bug narrowed to a race window.

The fix is one statement. Take a lock on the collection row before the preflight:

```ts
await tx
  .select({ id: schema.docCollection.id })
  .from(schema.docCollection)
  .where(and(
    eq(schema.docCollection.organizationId, principal.organizationId),
    eq(schema.docCollection.id, collectionId),
  ))
  .for('update');
```

That genuinely closes it, and the reason is worth putting in your commit message because no comment can carry it: `doc.collection_id` has a foreign key to `doc_collection.id`, so Postgres takes a `FOR KEY SHARE` lock on the collection row for any insert or update that places a document in it, and that conflicts with `FOR UPDATE`. I verified the block empirically, a concurrent insert waits for the whole transaction and only lands after commit.

**3. Batch the access check.** Greptile's P2 is fair. The loop runs one `writeGrantedDocIds` query per restricted document, sequentially, inside the transaction. `grantedDocIds` already batches exactly this shape, so a single lookup up front replaces N round trips. While you are there, `DOC_COLUMNS` pulls every document's full `content` body for a check that only needs `id`, `authorId`, `visibility` and `organizationId`.

**4. Put the guard inside `emptyCollection`.** The bulk `UPDATE` lives in `emptyCollection`, but the authorization lives in its caller. That is correct today because there is exactly one caller, but the invariant is not local to the function that does the damage, so the next caller reopens the hole silently. Moving the preflight and the lock into `emptyCollection` makes it hold for anyone who calls it.

**5. Pin the cases that must keep working.** Your test covers the refusal. Please also add one that an admin can still delete a collection holding another member's private document, and one that a published document does not block. Your implementation gets both right, but nothing currently fails if a future change breaks them, and they are exactly the ones that break quietly.

## Minor

- The test is named "refuses to move a doc the caller cannot write", but it deletes a collection. Worth renaming to match.

## One thing you should know

PR #316 is an independent attempt at this same issue, opened earlier and also targeting #245. The two touch the same region and conflict, so a maintainer will need to pick one rather than both landing. That is not a reflection on your work, and I would rather tell you now than after another round of changes. Your choice to go through `assertDocWritable` instead of re-implementing the rules is the better call of the two, for what it is worth.

imshashank:
## Decision on this one

We are consolidating #245 onto #316, which was opened about eleven hours before this PR for the same issue. Two people should not be spending evenings on the same bug, so I want to be direct rather than leave this sitting: **this PR will not be merged.**

That is a call about who picked the work up first, not about the quality of what you wrote. To be accurate about it: of the two implementations, yours is the better one in the way that matters most. You routed the decision through `assertDocWritable` instead of re-deriving the rules inline, and I verified that this is exactly why yours keeps the admin bypass working and correctly lets `public` and `link` documents through, while the other implementation currently breaks both. Your test asserting that the collection survives and the document placement is unchanged is also the better test, because it proves the rollback rather than just the rejection.

So this is not wasted. I am asking #316 to adopt both of those things by name, with a pointer back here.

The failure is ours: #245 was never assigned, and there was no note on it saying somebody had started. I have no way to give you those hours back, and I am sorry for them.

## To avoid this next time

Please comment on an issue to claim it before you start, and we will assign it to you. If an issue already has a comment claiming it, or a linked PR, treat it as taken. We will do a better job of keeping that state visible.

## Two things worth your time instead

Both are unclaimed, and I have checked that neither has an open PR against it:

- **#194, the audit log table is never written to.** Security, `area: database`, p1. `audit_log` ships on every deployment with a well designed set of columns and nothing has ever written a row to it, so an admin reading the schema concludes there is an audit trail that does not exist. This is the closest match to what you just demonstrated: service layer work, drizzle, transactions, and getting the authorization-adjacent details right. It is the larger of the two.
- **#191, send personal notifications as a Slack DM rather than to a channel.** Marked `ready`, p1, three points, and the routing rule is spelled out in the issue, so there is no design question waiting on me.

If you would rather have something smaller to land first, #183 and #175 are both small and marked `ready`.

One caution learned from #245: if an issue carries the `needs design` label, please ask me to make the call in the issue before writing code. #245 had it, and the ambiguity in that label is a direct cause of this collision. Neither #194 nor #191 has it.

The detailed technical review I left on this PR still stands and is worth reading, since the same points are going to land in #316. Thank you for the work, and I hope you will pick up one of the above.

imshashank:
Status update: per the decision above, this stays open only as a record. #316 is the PR carrying #245 forward, with your `assertDocWritable` routing and the rollback-style test folded in. Labeling this `duplicate`, no further action needed here.

Thanks again for the work, and I hope to see you on #194 or #191.

imshashank:
Closing this as a duplicate of #316, which came in first and has since landed the same fix with batched grant lookups instead of a query per document, a lock against concurrent deletes, and an error that names which pages are blocking. #316's CI is green and it is one round from approval, so consolidating there rather than carrying two branches for the same issue.

Not wasted work though, and I want to be specific about that rather than vague. Your test shape was better than what #316 had: asserting that the collection still exists and the document's `collectionId` is unchanged after the refusal, instead of only asserting that the call threw. That is the assertion that actually proves the transaction rolled back, and I asked #316 to adopt it, which it has. The verb agreement fix in the error message came across too.

Two things for next time, since the same pattern shows up in #321 and #323 as well:

- Check for an existing PR against the issue before starting. #316 was already open against #245 when this went up, and both of you spent a round on the same fix.
- The per document `assertDocWritable` loop here is N+1, one query per page. For a collection with a few hundred pages that is a few hundred round trips inside a write transaction. Batching the lookup, the way #316 does with a single `inArray`, is the habit worth taking from this.

Thanks for the quick turnaround on this one. If you want something else in the same area, #324 is open and unclaimed, and it happens to be the piece that unblocks your own Slack work on #321 and #323.

PR #331, opened 2026-08-17
feat(web): add shift+t keyboard shortcut for theme toggle

### LINKED ISSUE #130: The theme toggle has no keyboard shortcut
Every other command in the palette that has a keyboard equivalent declares a `binding`. The theme toggle does not, so switching theme means opening the palette and finding it.

**The fix**

In `apps/web/src/lib/navigation.ts`, the `view:theme` entry has no `binding` property:

```ts
{
  id: 'view:theme',
  label: context.dark ? 'Switch to light theme' : 'Switch to dark theme',
  section: 'View',
  icon: context.dark ? Sun : Moon,
  run: context.toggleTheme,
},
```

Compare it with the entry directly below, `view:sidebar`, which has `binding: '['`.

Adding a binding is enough. The shortcut then appears automatically in the `?` dialog, because that dialog is generated from the registry rather than from a separate list.

**Picking the key**

It has to not collide with anything already registered. The existing single-key global bindings are `?` and `[`, and `]` is registered in `apps/web/src/components/layout/app-shell.tsx`. Surface bindings such as `j`, `k`, `c`, `s`, `p`, `a`, `l`, `i`, `x`, `f`, `t`, `u`, `h` are taken on their own surfaces.

`shift+d` is probably free and reads as "dark". Suggest what you like on this thread before writing the code if you would rather check first.

**Also**

- Add a row to the View table in [docs/keyboard-shortcuts.md](https://github.com/Noveum/orbit/blob/main/docs/keyboard-shortcuts.md), which currently says the theme toggle has no key of its own
- A test in `apps/web/tests/` asserting the binding is registered

Read [CONTRIBUTING.md](https://github.com/Noveum/orbit/blob/main/CONTRIBUTING.md) first. The short version: `bun run verify` must be green, and there are no comments in code.

SinghAstra:
Resolves #130. I noticed this issue had been open without a linked PR for a week, so I went ahead and put the fix together to help clear out the backlog!

Binds shift+d to the view:theme command in navigation.ts.

Updates the View table in docs/keyboard-shortcuts.md.

Adds a unit test in navigation.test.ts asserting the binding.

bun test apps/web/tests/lib/navigation.test.ts is fully green.

<h3>Greptile Summary</h3>

The PR adds a global Shift+T shortcut for toggling the theme and keeps its test coverage and keyboard-shortcut documentation aligned.
- Registers `shift+t` on the existing `view:theme` command.
- Adds a unit test asserting the command binding.
- Documents the new shortcut and refreshes the shortcut-registration example.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| apps/web/src/lib/navigation.ts | Adds the supported and currently unclaimed `shift+t` binding to the existing theme command. |
| apps/web/tests/lib/navigation.test.ts | Adds focused coverage confirming that the theme command exposes the new binding. |
| docs/keyboard-shortcuts.md | Documents Shift+T in the View shortcuts table and updates the illustrative registration example. |

</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into feat/130-theme-..."](https://github.com/noveum/orbit/commit/90f3e457c1e74e1011ae42efbf68c405b050ad64) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53932517)</sub>

**Context used:**

- Knowledge Base — [Orbit web application](https://app.greptile.com/noveum-ai/-/custom-context/knowledge-base/noveum/orbit/-/docs/web-application.md)

### DIFF
--- a/apps/web/src/lib/navigation.ts
+++ b/apps/web/src/lib/navigation.ts
@@ -208,6 +208,7 @@ export function buildCommands(context: CommandContext): AppCommand[] {
       label: context.dark ? 'Switch to light theme' : 'Switch to dark theme',
       section: 'View',
       icon: context.dark ? Sun : Moon,
+      binding: 'shift+t',
       run: context.toggleTheme,
     },
     {
--- a/apps/web/tests/lib/navigation.test.ts
+++ b/apps/web/tests/lib/navigation.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'bun:test';
-import type { NavLink, ShellTeam } from '@/lib/navigation.ts';
-import { buildNavigation, buildSidebarNav } from '@/lib/navigation.ts';
+import type { CommandContext, NavLink, ShellTeam } from '@/lib/navigation.ts';
+import { buildCommands, buildNavigation, buildSidebarNav } from '@/lib/navigation.ts';
 
 const TEAMS: readonly ShellTeam[] = [
   { id: 'team_1', key: 'ENG', name: 'Engineering', openIssues: 4 },
@@ -55,5 +55,22 @@ describe('navigation bindings', () => {
     expect(buildSidebarNav(TEAMS, 7).personal[0]?.count).toBe(7);
     expect(fromSections?.count).toBe(2);
     expect(fromSidebar?.count).toBe(7);
+  });
+});
+
+describe('global commands', () => {
+  it('registers the theme toggle with the shift+t binding', () => {
+    const context: CommandContext = {
+      sections: [],
+      navigate: () => undefined,
+      toggleSidebar: () => undefined,
+      toggleTheme: () => undefined,
+      showShortcuts: () => undefined,
+      dark: false,
+    };
+    const commands = buildCommands(context);
+    const themeCommand = commands.find((cmd) => cmd.id === 'view:theme');
+    expect(themeCommand).toBeDefined();
+    expect(themeCommand?.binding).toBe('shift+t');
   });
 });
--- a/docs/keyboard-shortcuts.md
+++ b/docs/keyboard-shortcuts.md
@@ -117,8 +117,7 @@ contributors and guests do not.
 | <kbd>Cmd</kbd> <kbd>B</kbd> | Switch between board and list |
 | <kbd>[</kbd> | Toggle the left sidebar |
 | <kbd>]</kbd> | Toggle the right panel |
-
-The theme toggle has no key of its own. Run it from the command palette.
+| <kbd>Shift</kbd> <kbd>T</kbd> | Switch between light and dark theme |
 
 ## How the shortcuts resolve
 
@@ -147,7 +146,7 @@ what makes a shortcut appear in the <kbd>?</kbd> dialog. There is no separate
 list to update:
 
 ```ts
-useHotkey('shift+d', duplicateIssue, {
+useHotkey('shift+z', duplicateIssue, {
   label: 'Duplicate issue',
   section: 'Issues',
   scope: 'issues',


apps/web/src/lib/navigation.ts
```diff
@@ -208,6 +208,7 @@ export function buildCommands(context: CommandContext): AppCommand[] {
       label: context.dark ? 'Switch to light theme' : 'Switch to dark theme',
       section: 'View',
       icon: context.dark ? Sun : Moon,
+      binding: 'shift+d',
```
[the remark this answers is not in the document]
SinghAstra: Thanks for the note! Checked the keyboard shortcut registry and documentation (docs/keyboard-shortcuts.md)—there is currently no shortcut bound to Shift+D or d for due dates (g d is used for navigation to Docs, but that's a two-key prefix). Shift+D remains free and aligns directly with the issue description.

imshashank:
@coderabbitai review

apps/web/src/lib/navigation.ts
imshashank: This is the collision. `issue-properties.tsx:107` already registers `shift+d` for the due date menu, and `issue-properties.tsx:277` advertises it in the UI as the due date shortcut.

Both register at `HOTKEY_PRIORITY.global`, because the issue one sets `scope: 'issues'` but never sets `priority`, and scope does not participate in `beats()`. With equal step count, equal priority and equal shift specificity, `beats()` returns false and `selectMatch` keeps the first entry it encountered. The palette registers on shell mount, before any issue page, so the theme command wins and the due date shortcut silently stops opening its menu.

I confirmed this against the real `HotkeyRegistry` rather than reading it off the rules: registering both and calling `selectMatch` returns `Switch theme / palette-theme`.

`shift+t` is free and mnemonic. Changing this one line and the docs row fixes it.

Worth noting for future work: the registry silently resolves a tie rather than complaining about it, which is how this got past review. A duplicate binding check in the registry, or a test that asserts no two advertised entries share a binding, would catch the next one automatically. That is out of scope here, and I will open an issue for it.

apps/web/tests/lib/navigation.test.ts
```diff
@@ -57,3 +57,20 @@ describe('navigation bindings', () => {
...
+      navigate: () => undefined,
+      toggleSidebar: () => undefined,
+      toggleTheme: () => undefined,
+      showShortcuts: () => undefined,
+      dark: false,
+    };
+    const commands = buildCommands(context);
+    const themeCommand = commands.find((cmd) => cmd.id === 'view:theme');
+    expect(themeCommand).toBeDefined();
+    expect(themeCommand?.binding).toBe('shift+d');
```
imshashank: This asserts the metadata rather than the behaviour, so it passes whether or not the shortcut actually works. It would have passed unchanged with the collision in place, which is the definition of a test that cannot fail for the thing it is meant to protect.

What would earn its keep: build a `HotkeyRegistry`, register the palette command alongside a `scope: 'issues'` entry on the same key, and assert `selectMatch` picks the one you intend. That is the assertion that breaks the day somebody adds a third `shift+t`.

The fixture construction here is good, and `CommandContext` being explicit rather than cast is the right instinct.

docs/keyboard-shortcuts.md
```diff
@@ -116,9 +116,8 @@ contributors and guests do not.
 | --- | --- |
 | <kbd>Cmd</kbd> <kbd>B</kbd> | Switch between board and list |
 | <kbd>[</kbd> | Toggle the left sidebar |
-| <kbd>]</kbd> | Toggle the right panel |
-
-The theme toggle has no key of its own. Run it from the command palette.
+| <kbd>]</kbd> | Toggle the right panel  |
+| <kbd>Shift</kbd> <kbd>D</kbd> | Switch between light and dark theme |
```
imshashank: Two things in this file.

The row is correct once the binding moves, so it just needs `Shift` `T` instead.

More importantly, this file now contradicts itself. Line 149, in the "Adding one" section, uses `useHotkey('shift+d', duplicateIssue, ...)` as the worked example for registering a shortcut. So the same document tells a reader that Shift+D is the theme toggle and, thirty lines later, that it is duplicate issue. Whichever key this lands on, please pick a key for that example that is not bound to anything real, so the example cannot go stale again.

Also a stray edit: line 119 picked up two trailing spaces after `Toggle the right panel`. Worth dropping before merge, since trailing whitespace in a table cell is invisible in review and shows up in the next diff that touches the line.

imshashank (changes requested):
Thanks for picking this up unprompted. #130 had sat for a week and you found it and fixed it without being asked, which is exactly the contribution pattern I want, and the shape of the change is right: one binding, one doc update, one test.

There is one real problem, and it is not quite the one Greptile described.

## Shift+D is already taken, and the theme toggle wins

`apps/web/src/features/issues/issue-properties.tsx:107` already binds `shift+d` to the due date menu, and the UI advertises it: `<PropertyRow label="Due date" shortcut="shift+d">`.

Greptile said the theme shortcut would not reliably fire on issue pages. I checked against the real registry and it is the other way round, which is worse. Both entries register at `priority: 0`. The issue one passes `scope: 'issues'` but no priority, so it falls back to `HOTKEY_PRIORITY.global`. In `beats()` the two are identical on all three tiebreakers: one step each, same priority, one shift each. So `beats` returns false, `selectMatch` keeps whichever it saw first, and that is the command palette's entry, because the palette mounts with the shell before any issue page exists.

I ran it against `HotkeyRegistry` and `selectMatch` directly rather than reasoning about it:

```
WINNER: Switch theme / palette-theme
```

So after this change, pressing Shift+D on an issue detail page toggles the theme and the due date menu no longer opens. An existing, advertised, documented shortcut stops working, and nothing in the test suite notices.

## Suggested fix

Use `shift+t`. I inventoried every binding in the app and it is free:

```
useHotkey:  ] a alt+shift+f alt+v escape f i j k l m p s shift+d shift+e shift+f shift+j shift+k
palette:    ? [ g a g d g i g m g p g r g s g t g v
aliases:    up down shift+up shift+down
```

It is mnemonic for theme, it does not collide, and `g t` is a two step sequence so it does not interfere. Swap the binding, update the docs row, and this is done.

If you would rather keep `shift+d` for the theme, that is a bigger conversation: it would mean deliberately taking a key off the issue surface and rebinding due date, and I would want that argued on its own rather than as a side effect.

## Smaller things

Three minor items inline. None of them are blocking on their own.

Once the binding changes, this is a quick approve from me. Sorry you waited a day on it, the answer turned out to be one word.

imshashank:
Opened #332 for the underlying gap, since the registry resolving a tie silently is what let this through review rather than anything you did. Tagged `good first issue` if you want it after this one.

SinghAstra:
Pushed the fix!

1. Changed the binding to shift+t in navigation.ts and navigation.test.ts.

2. Updated the View table to Shift+T in the docs.

3. Changed the duplicateIssue example key to shift+z in the docs so it doesn't drift.

4. Dropped the trailing whitespace on line 119.

Good catch on selectMatch! I see you opened #332 for the underlying architecture fix, so I kept this PR strictly to the Theme toggle shortcut to stay in scope.

imshashank (approved):
All three addressed, and the docs example fix is the one I care most about: moving that snippet to `shift+z` means the worked example is no longer describing a key that is really bound to something else, so it cannot drift into a real conflict again. I checked `shift+z` against the full binding inventory and it is free.

Verified locally: `shift+t` does not collide with anything, `apps/web` typechecks clean, and `tests/lib/navigation.test.ts` is 6 pass 0 fail.

On my second comment, the one asking for a registry level test rather than a metadata assertion: you have effectively answered it in #334 by testing the registry directly, which is the better place for it. Keeping the metadata assertion here is fine, since what this PR owns is the choice of key rather than the resolution behaviour.

Thanks for turning this around quickly, and for finding the issue yourself in the first place.

imshashank:
Green now. The `Unit and integration tests` failure you saw here was never yours: #323 had added a test that set `BETTER_AUTH_SECRET` at module scope, which changed it for the whole test process and broke MCP token verification in `MCP authorize PKCE boundary`. The Playwright job then aborted as a knock-on rather than failing on its own. Fixed in #338 and `main` is green again.

Merging this. Thanks for the fix and for the follow up in #334.

PR #334, opened 2026-08-18
feat(web): add development collision detection for hotkeys

### LINKED ISSUE #332: Two shortcuts on the same key silently resolve instead of failing
Found while reviewing #331, which bound `shift+d` to the theme toggle without noticing that `issue-properties.tsx` already binds it to the due date menu.

## What happens

`selectMatch` in `apps/web/src/lib/keyboard/registry.ts` resolves a pressed binding by walking every registered entry and keeping the best one, where `beats()` compares step count, then priority, then shift specificity. When two entries tie on all three, `beats()` returns false and the winner is simply whichever entry the registry happened to see first.

That is a silent resolution, not an error. Nothing warns, nothing fails, and the losing shortcut just stops working. In #331's case the command palette registers on shell mount, before any issue page, so the global theme toggle would have shadowed a surface shortcut that the UI actively advertises through `<PropertyRow shortcut="shift+d">`.

It is also invisible to review. The PR was small, well tested and correctly documented, and the collision was still only caught by reading every other `useHotkey` call in the app.

## What to build

A duplicate detection that fails rather than resolves. Two reasonable shapes, either is fine:

- A test that builds the full registry the way the app does and asserts no two advertised entries share a binding within a scope that can be active at the same time.
- A development mode warning from `HotkeyRegistry.register` when an incoming entry ties an existing one on all three tiebreakers, since a genuine tie means neither author knew about the other.

The test is probably the better first step because it runs in CI and cannot be ignored.

## Also worth doing

`docs/keyboard-shortcuts.md` uses `useHotkey('shift+d', duplicateIssue, ...)` as its worked example in the "Adding one" section, while `shift+d` is really the due date menu. Pick a key for that example that is not bound to anything, so the documentation cannot drift into describing a real conflict.

Related: #130, #331.

SinghAstra:
## What this changes

Adds a development-mode warning check inside `HotkeyRegistry.register()` to detect and flag silent hotkey tie collisions, along with a dedicated unit test suite.

## Why

When two advertised shortcuts tie on all three tiebreakers (steps, shift specificity, and priority), the registry silently resolves them based on registration order. This caused issues like #331 where the theme toggle shadowed the due-date menu shortcut. This check ensures developers are loudly warned in non-production environments if a tie occurs. 

Closes #332

## How you know it works

Added a new unit test file at `apps/web/tests/lib/keyboard/registry.test.ts`. 

Ran `bun test apps/web/tests/lib/keyboard/registry.test.ts` to verify that:
1. Identical priorities triggering a duplicate binding correctly fire a `console.warn`.
2. Differing priorities (global vs surface) pass cleanly without warnings.

## Screenshots

*Not applicable.*

## Checklist

- [ ] `bun run verify` is green, all four checks
- [ ] No comments added to code, and no em-dash characters anywhere
- [ ] External input is parsed with a Zod schema from `@orbit/shared`
- [ ] Authorization is enforced on the server through `packages/shared/src/policy`, not only in the UI

Implemented option 2 from the issue description: a non-production `console.warn` inside `HotkeyRegistry.register()` when advertised hotkeys tie on all tiebreakers, allowing existing integration/E2E test runs to surface collisions.

<h3>Greptile Summary</h3>

The PR adds development-only warnings when advertised hotkeys tie on binding structure, Shift specificity, and priority, plus focused registry tests.
- Detects collisions across identical and differently positioned Shift modifiers.
- Avoids warnings where priority resolves the conflict.
- Restores test-level console stubs through exception-safe cleanup.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| apps/web/src/lib/keyboard/registry.ts | Adds collision detection whose comparison now matches the registry's runtime matching and tie-break semantics. |
| apps/web/tests/lib/keyboard/registry.test.ts | Covers duplicate, priority-separated, and multi-step Shift collisions while restoring console.warn in finally blocks. |

</details>

<sub>Reviews (3): Last reviewed commit: ["feat(web): add development collision det..."](https://github.com/noveum/orbit/commit/ac0b21f26b207feba24f1c679e306be6df795fcd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54427382)</sub>

### DIFF
--- a/apps/web/src/lib/keyboard/registry.ts
+++ b/apps/web/src/lib/keyboard/registry.ts
@@ -32,6 +32,23 @@ function shiftSpecificity(entry: HotkeyEntry): number {
   return entry.steps.reduce((total, step) => total + (step.shift ? 1 : 0), 0);
 }
 
+function stepsCollide(a: readonly HotkeyStep[], b: readonly HotkeyStep[]): boolean {
+  if (a.length !== b.length) return false;
+  let shiftA = 0;
+  let shiftB = 0;
+  for (let i = 0; i < a.length; i++) {
+    const stepA = a[i];
+    const stepB = b[i];
+    if (!(stepA && stepB)) return false;
+    if (stepA.key !== stepB.key || stepA.mod !== stepB.mod || stepA.alt !== stepB.alt) {
+      return false;
+    }
+    if (stepA.shift) shiftA++;
+    if (stepB.shift) shiftB++;
+  }
+  return shiftA === shiftB;
+}
+
 function beats(candidate: HotkeyEntry, current: HotkeyEntry): boolean {
   if (candidate.steps.length !== current.steps.length) {
     return candidate.steps.length > current.steps.length;
@@ -61,7 +78,23 @@ export class HotkeyRegistry {
   private snapshot: readonly HotkeyEntry[] = [];
 
   register(input: HotkeyEntryInput): () => void {
-    this.entries.set(input.id, { ...input, steps: parseBinding(input.binding) });
+    const steps = parseBinding(input.binding);
+
+    if (process.env.NODE_ENV !== 'production' && input.advertised) {
+      for (const existing of this.entries.values()) {
+        if (existing.id === input.id) continue;
+        if (
+          existing.advertised &&
+          existing.priority === input.priority &&
+          stepsCollide(existing.steps, steps)
+        ) {
+          console.warn(
+            `Duplicate hotkey collision: "${input.binding}" is advertised by both "${existing.id}" and "${input.id}" at priority ${input.priority}. This will cause silent resolution.`,
+          );
+        }
+      }
+    }
+    this.entries.set(input.id, { ...input, steps });
     this.publish();
     return () => {
       this.entries.delete(input.id);
--- a/apps/web/tests/lib/keyboard/registry.test.ts
+++ b/apps/web/tests/lib/keyboard/registry.test.ts
@@ -1,0 +1,82 @@
+import { expect, test } from 'bun:test';
+import {
+  HOTKEY_PRIORITY,
+  type HotkeyEntryInput,
+  HotkeyRegistry,
+} from '../../../src/lib/keyboard/registry.ts';
+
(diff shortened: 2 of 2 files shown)


imshashank (approved):
This is the right shape and it would have caught #331: two advertised entries, same key, same priority, warned about instead of silently resolved. Test file is in the correct mirrored path, the `process.env.NODE_ENV` guard keeps it out of production bundles, and skipping the self-comparison by `id` means a re-register does not warn at itself. 47 tests in `apps/web/tests/lib/keyboard/` pass locally.

I want to flag the part I got wrong while reviewing, because it is the most interesting thing in the PR.

I had written you a comment arguing that `stepsCollide` was buggy: that ignoring `shift` per step and comparing the totals at the end was too loose, that `shift+a b` and `a shift+b` can never both match one keystroke sequence, and that your third test was therefore asserting a false positive. I was about to send it. Then I checked `stepMatches` before posting:

```ts
return step.shift ? actual.shift : true;
```

A step without `shift` matches a shifted press. So pressing Shift+A then Shift+B matches **both** bindings. I proved it rather than reasoning about it a second time:

```
shift+a b matches Shift+A,Shift+B : true
a shift+b matches Shift+A,Shift+B : true
```

Both match, both have two steps, both have shift specificity 1, so `beats()` ties and `selectMatch` silently picks whichever registered first. That is exactly the class of bug this PR exists to catch, and comparing shift totals rather than per step flags is what catches it. Your version is correct and my correction would have introduced a false negative that quietly stopped detecting real collisions.

Worth a line in the PR description or the issue, because the next person to read `stepsCollide` will have the same objection I did, and the reason it is written that way is not obvious from the function alone.

Two small things, neither blocking:

- The test imports `'../../../src/lib/keyboard/registry.ts'`. Everything else under `apps/web` uses the `@/` alias and `AGENTS.md` asks for it. `@/lib/keyboard/registry.ts` also survives the file moving.
- The rest of the suite uses `describe` with `it`, this file uses bare `test`.

Happy to take those in a follow up rather than another round here. Approving.

PR #335, opened 2026-08-18
test(e2e): fix account and workspaces flaky test

### LINKED ISSUE #255: account-and-workspaces e2e flakes on a duplicated provider card
Failed once on main (`373ca37`, run 31274006508), passed on re-run of the same commit, and passed on a branch containing that same commit. Pre-existing and intermittent, not a regression from anything merged around it.

## What Playwright reported

```
Error: strict mode violation: getByTestId('provider-github').getByText('Connected')
resolved to 2 elements:
  2 × <span class="...text-success">Connected</span>
      - unexpected value "hidden"
```

`apps/web/e2e/account-and-workspaces.spec.ts:85`

## What it is not

`connected-accounts.tsx` renders exactly one `<Badge tone="success">Connected</Badge>` per provider card, and `describeProvider` returns `Account <id>` when linked, so the description does not contain the word. Two identical badges means `getByTestId('provider-github')` matched **two list items**, not that one card held two badges.

## Likely cause

One of the two is hidden, which is what a soft navigation looks like while the outgoing tree is still mounted: the test does `page.goto('/settings/account/connections')` and asserts immediately, with no wait for the previous view to unmount. Playwright resolves both trees and strict mode rejects the ambiguity.

That would also explain why it is rare and why a re-run passes.

## Worth fixing regardless

Even if the duplication is transient, the assertion is too loose to be trustworthy: it passes or fails depending on which tree Playwright resolves first. Options, roughly in order of preference:

1. wait for the connections view to settle before asserting, so only one tree is mounted
2. scope to the visible card, e.g. `getByTestId('provider-github').filter({ visible: true })`
3. give the badge its own test id and assert on that

Reproducing it reliably probably needs the e2e stack running in a loop, since it did not reproduce on demand.

SinghAstra:
## What this changes

Fixes an intermittent strict mode violation in the `account-and-workspaces.spec.ts` end-to-end test.

## Why

As outlined in the issue, when Playwright navigates to `/settings/account/connections`, Next.js begins rendering the incoming page but hasn't yet unmounted the outgoing tree. Playwright resolves both trees simultaneously, finds two elements matching `getByTestId('provider-github')`, and crashes due to a strict mode ambiguity.

Closes #255

## How you know it works

Added `.filter({ visible: true })` to both occurrences of the `provider-github` locator in the test. 

To be completely honest, I didn't run the test command locally. I ran into a local Windows environment issue where the Playwright workers weren't picking up my `.env` variables (`DATABASE_URL`). Rather than fighting my local setup, I pushed the fix directly to let the GitHub CI pipeline verify the E2E run. 

## Screenshots

*Not applicable.*

## Checklist

- [ ] `bun run verify` is green, all four checks *(relying on CI)*
- [ ] No comments added to code, and no em-dash characters anywhere
- [ ] External input is parsed with a Zod schema from `@orbit/shared`
- [ ] Authorization is enforced on the server through `packages/shared/src/policy`, not only in the UI

Went with the suggested `.filter({ visible: true })` approach since it requires no changes to the actual React component and directly addresses the Playwright DOM resolution issue.

<h3>Greptile Summary</h3>

Updates the account and workspace end-to-end test to disambiguate duplicate GitHub provider cards during page transitions.
- Filters both `provider-github` locators to visible elements.
- Preserves the existing connected-state and disconnect-button assertions.

The PR appears safe to merge with no actionable defects identified.

The pinned Playwright version supports the visibility filter, and the locator narrowing is consistent with the duplicate outgoing and incoming tree scenario the test is intended to handle.

| Filename | Overview |
|----------|----------|
| apps/web/e2e/account-and-workspaces.spec.ts | The updated locators use a Playwright-supported visibility filter to avoid strict-mode ambiguity without changing test intent. |

</details>

<sub>Reviews (1): Last reviewed commit: ["test(e2e): fix account and workspaces fl..."](https://github.com/noveum/orbit/commit/bc3dfba879c3748b9374430cb6cdb9782a6a4127) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54437313)</sub>

### DIFF
--- a/apps/web/e2e/account-and-workspaces.spec.ts
+++ b/apps/web/e2e/account-and-workspaces.spec.ts
@@ -81,7 +81,7 @@ test('account settings, passkeys, and workspace switching', async ({ browser }) 
   }
 
   await page.goto(`${BASE}/settings/account/connections`);
-  const github = page.getByTestId('provider-github');
+  const github = page.getByTestId('provider-github').filter({ visible: true });
   await expect(github.getByText('Connected')).toBeVisible();
   await expect(github.getByRole('button', { name: 'Disconnect' })).toBeDisabled();
   await expect(
@@ -107,7 +107,10 @@ test('account settings, passkeys, and workspace switching', async ({ browser }) 
 
   await page.goto(`${BASE}/settings/account/connections`);
   await expect(
-    page.getByTestId('provider-github').getByRole('button', { name: 'Disconnect' }),
+    page
+      .getByTestId('provider-github')
+      .filter({ visible: true })
+      .getByRole('button', { name: 'Disconnect' }),
   ).toBeEnabled();
   await page.screenshot({
     path: `${SHOTS}/04-connections-unlockable.png`,


imshashank (approved):
This matches the diagnosis in #255 exactly. The issue concluded that two trees are briefly mounted during the soft navigation and one is hidden, and `.filter({ visible: true })` is the right way to say "the one the user can actually see" rather than widening the assertion or adding a sleep.

Checked the things that could have bitten:

- `filter({ visible: true })` needs Playwright 1.51 or newer, and `apps/web` is on `^1.62.1`.
- Both `getByTestId('provider-github')` call sites in the spec are covered. I grepped, lines 84 and 110 were the only two, and you got both.
- The e2e job passed on CI.

Thanks for saying plainly that you did not run it locally and leaned on CI instead. That is the right call for an e2e-only change and much more useful to me than a checkbox I would have had to take on trust.

Issue #219, opened 2026-08-08
Audit and widen the MCP tool surface


imshashank:
The MCP server covers a lot. `packages/mcp-server/src/tools/` has eleven files spanning issues, docs, planning, scrum, taxonomy, workspace, org, identity, admin and support.

It does not cover everything, and the gaps are not obvious until an agent hits one mid-task and has no way forward.

## What this issue is

An audit and a fill. It is deliberately open ended and it suits somebody who wants to use Orbit through an agent and fix what they run into.

**Step one, which is a contribution on its own:** connect an agent, try to do a realistic piece of work end to end, and write down every point where a tool was missing, took the wrong shape, returned too much, returned too little, or had a description the model misread. Post that list here. Somebody else can take the fixes.

## Known gaps

Some are already visible from reading the code:

- **Analytics.** There is no analytics tool at all. `buildChart` is a well defined function with a Zod request schema at `packages/core/src/analytics/schemas.ts` and it would map to a tool almost directly. An agent that can read the board but cannot answer "how did the team do last sprint" is missing the most natural question to ask it
- **Anything behind a dormant table.** Webhooks, automation rules, the audit log and API keys have no tools because they have no features. They arrive with those features
- **Attachments and files.** Check what exists
- **Notification and inbox state.** An agent that can triage should be able to mark things read
- **Bulk operations.** Moving forty issues one call at a time is slow and hits rate limits. Where a batch is safe, offer one

## Quality, not just coverage

More tools is not automatically better. Past a certain number, tool descriptions start competing for the model's attention and the whole set gets worse. So:

- **Descriptions are the interface.** A tool whose description does not say when to use it will be used wrongly. Read the existing ones for the standard
- **Names over ids.** The server instructions already say tools take names rather than ids, which is right for a model. Keep it
- **Errors are instructions.** A failure should tell the model what to do differently, not just that something went wrong. "No team named Platform. Call list_teams to see the options." is a good error
- **Response size.** A tool returning two hundred issues in full burns context for no benefit. Paginate and summarise
- **Scopes.** Every read tool needs `orbit.read`, every write tool needs `orbit.write`, and a token with neither is refused before any tool is registered. Any new tool follows that

## Tests

`packages/mcp-server/` has its own suite and its own test database, `orbit_test_mcp`. Every new tool gets a test, including a permission test proving a member cannot reach another team through it.

## Files

- `packages/mcp-server/src/tools/`
- `packages/mcp-server/src/server.ts`
[...]

Issue #317, opened 2026-08-15
Analytics lets a guest read issues on teams they are not a member of


imshashank:
Found while triaging #178. This is a question about an intentional decision rather than a report of a mistake, so it needs a yes or no from a maintainer before anybody writes code.

## What the code does

`visibilityFilters` in `packages/core/src/work/issue-query.ts` has two modes:

```ts
function visibilityFilters(principal: Principal, visibility: IssueVisibility): SQL[] {
  if (visibility === 'workspace-analytics') {
    assertCan(principal, 'analytics:read');
    return [];
  }
  assertCan(principal, 'issue:read');
  return visibleTeamFilters(principal);
}
```

`workspace-analytics` applies no team filter. The only caller is `baseAnalyticsPredicate` in `packages/core/src/analytics/drilldown.ts`, which every lens and the drilldown go through.

`analytics:read` is the first entry in `GUEST_PERMISSIONS` in `packages/shared/src/policy/index.ts`, so every role in the workspace holds it, guest included.

## Why it is worth confirming

Three things make this larger than a scoping choice:

1. **It reverses the previous behaviour.** The engine this replaced restricted a non-admin to their own teams. `visibleTeamsClause` in `packages/core/src/analytics/distribution.ts` still shows it: `and i.team_id in (...)` for a member, and `and false` for somebody on no teams. The planning cockpit in #309 moved the shipping path onto `workspace-analytics`, and the old clause is now only reachable from unused code.

2. **The drilldown returns rows, not just totals.** `listAnalyticsDrilldown` selects `issue.identifier`, `issue.title`, priority, estimate, due date and state name. So this is not "a guest sees an aggregate that includes other teams", it is "a guest can enumerate the identifiers and titles of issues on a team they cannot open in the issue list". The same person gets nothing for those issues through the `team` visibility that the issue list uses.

3. **#178 asked for the opposite in writing.** It specified a test that "a member of team A cannot request a scope of team B and get numbers back". The implementation is the reverse of that, so the divergence should be recorded on purpose rather than discovered later.

## It is deliberate, and pinned by a test

`packages/core/tests/work/issue-query.test.ts` asserts it directly. A guest, contributor and member each on only the engineering team see both `Engineering work` and `Operations work` through `workspace-analytics`, and only `Engineering work` through `team`. So somebody chose this and wrote the test that keeps it that way. Nothing here is accidental, which is exactly why it needs a decision rather than a patch.

## The options

[...]

PR #336, opened 2026-08-18
feat(mcp): add analytics overview tool for agent insights (#219)

SinghAstra:
## What this changes

Exposes the core analytics overview engine as the read-only MCP tool `get_analytics_overview`.

The tool:

- reports stable metric IDs for throughput, WIP, cycle time, blocked, overdue, stale, and unestimated work
- accepts workspace-safe date presets and issue or points measurement
- returns effective ISO date and comparison ranges with timezone
- omits raw issue outliers and reports their total withheld count
- enforces the existing MCP `orbit.read` scope and core authorization policy

## Why

Addresses #219. Agents using Orbit through MCP need a safe, stable overview of workspace delivery health without receiving issue-level evidence they may not be allowed to inspect.

## How you know it works

Validated exact head `f8f542a9` after merging current `main`:

- core analytics overview: 12 tests, 111 assertions
- complete MCP server package: 199 tests, 932 assertions
- focused MCP analytics contract: 5 tests, 60 assertions
- `bun run typecheck`
- `bun run lint`
- `bun run check-comments`
- `git diff --check`

The complete hosted Actions gate is green. A full local `bun run verify` passed static checks, types, database, core, services, realtime, and MCP before Bun 1.3.14 hit a cumulative runner `SIGTRAP` at the web line-plot file; that file passes alone with 16 tests and 77 assertions.

The tests cover scope registration, invalid sprint presets, defaults, stable metric semantics, explicit 90-day points semantics, prior-period boundaries, effective range serialization, and withholding both visible and permission-restricted outliers.

## Checklist

- [x] Current `main` is merged
- [x] Tests added or updated, with mutation-sensitive assertions
- [x] No comments added to code and no em dash characters
- [x] No `any` and no non-null assertions
- [x] External input is parsed through the shared analytics Zod schema
- [x] Authorization is enforced on the server
- [x] No database migration is required
- [x] Exact-head hosted checks and bot reviews are complete

<h3>Greptile Summary</h3>

The PR exposes the existing analytics overview engine through a read-only MCP tool and adds wire-safe resolved date ranges to the core response.
- Registers `get_analytics_overview` for callers with read access.
- Validates range and measure inputs before translating them into the core analytics query.
- Returns projected overview metrics while withholding raw issue outliers.
- Adds core and MCP integration coverage for ranges, authorization, defaults, translation, and response projection.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/core/src/analytics/overview.ts | Adds ISO-serialized resolved and comparison ranges to the analytics overview result. |
| packages/core/tests/analytics/overview.test.ts | Verifies that effective analytics ranges are returned as wire-safe values. |
| packages/mcp-server/src/tools/analytics.ts | Defines the read-only analytics overview tool, validates inputs, and projects a privacy-conscious response. |
| packages/mcp-server/src/tools/index.ts | Registers the analytics tool with the MCP tool catalog. |
| packages/mcp-server/tests/tools/analytics.test.ts | Covers registration, scope filtering, defaults, input translation, response projection, and outlier withholding through the MCP dispatch path. |

</details>

```mermaid
sequenceDiagram
  participant C as MCP client
  participant S as MCP server
  participant T as Analytics tool
  participant A as Core analytics
  C->>S: get_analytics_overview(range, measure)
  S->>T: Validate scope and arguments
  T->>A: loadAnalyticsOverview(principal, query)
  A-->>T: Overview and resolved ranges
  T-->>S: Project metrics and withheld count
  S-->>C: Structured analytics response
```
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/mai..."](https://github.com/noveum/orbit/commit/f8f542a9cb1c06dd9fe15f9dd84fcd01965d9d17) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54454820)</sub>

**Context used:**

- Knowledge Base - [Analytics and planning insights](https://app.greptile.com/noveum-ai/-/custom-context/knowledge-base/noveum/orbit/-/docs/analytics-and-insights.md)
- Knowledge Base - [Model Context Protocol interface](https://app.greptile.com/noveum-ai/-/custom-context/knowledge-base/noveum/orbit/-/docs/mcp-interface.md)


packages/mcp-server/src/tools/index.ts
```diff
@@ -15,6 +16,7 @@ import { registerWorkspaceTools } from './workspace.ts';
 export function registerTools(server: McpServer, principal: Principal): void {
   registerIdentityTools(server, principal);
   registerInboxTools(server, principal);
+  registerAnalyticsTools(server, principal);
```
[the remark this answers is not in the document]
SinghAstra: Thanks for the review! Note that packages/mcp-server/src/tools/ does not use per-tool unit test files (e.g., inbox.ts and issues.ts have no corresponding .test.ts files). Tool registration and schemas are validated via TypeScript and integrated at the server dispatch layer.

imshashank (changes requested):
The tool itself is put together well. `defineTool` with `readOnly: true`, the input schema mirrors `analyticsQuerySchema` rather than inventing a second vocabulary, the descriptions are written for an agent to choose between, and reshaping `cards` and `outliers` into flat metric objects is the right instinct for something a model has to read. `packages/mcp-server` is 193 pass 0 fail with it in.

I cannot merge it, for a reason that is nothing to do with the code you wrote.

## It exposes an open security question

#317 is open, labelled `security` and `p1 high`, and it says this:

> `visibilityFilters` in `packages/core/src/work/issue-query.ts` has two modes. `workspace-analytics` applies no team filter. The only caller is `baseAnalyticsPredicate` in `packages/core/src/analytics/drilldown.ts`, which every lens and the drilldown go through. `analytics:read` is the first entry in `GUEST_PERMISSIONS`, so every role in the workspace holds it.

`loadAnalyticsOverview` goes through that path. So today, in the web app, a guest can already read analytics spanning teams they are not a member of. That is the open question, and #317 explicitly says it needs a yes or no from a maintainer before anybody writes code against it.

This PR takes that surface and hands it to agents over MCP. `readOnly: true` means the tool is registered for any token carrying `orbit.read`, so a guest's OAuth grant reaches it. And the `outliers` array returns issue identifiers and titles, which is the concrete leak rather than an aggregate number.

So merging this would widen an unresolved p1 into a second channel before the decision behind it is made. That ordering has to go the other way. I will get #317 decided, and once the team filter question is settled this becomes an easy merge.

Not something you could have known without reading #317, and I should have linked it on #219 when I labelled that one `good first issue`. That is on me.

## It also needs tests

Separately, and this is the `needs tests` label rather than the blocker: `packages/mcp-server/tests/` has `tools.test.ts`, `tools-smoke.test.ts` and `view-tools.test.ts`, and nothing covers the new tool. `bun run verify` passes only because the smoke test asserts `SMOKE.length > 15` rather than an exact set, so an untested tool slips through.

Worth covering:

1. The tool is registered when the token carries `orbit.read`, and refused when it does not.
2. `range` and `measure` map onto `analyticsQuerySchema` as intended, including the `auto` default.
3. The returned shape, so a change to `overview.cards` cannot silently alter the agent-facing contract.

Number 1 matters most, and it is also the test that will encode whatever #317 concludes.

## One smaller thing

`registerAnalyticsTools` sits between inbox and issues in `registerTools`. The surrounding order looks deliberate. Worth matching it or saying why.

Sorry to park this one when the code is fine. It is genuinely a sequencing problem rather than a quality one.

SinghAstra:
Totally makes sense, Shashank! Security always comes first. I hadn't seen #317, but I completely agree that exposing issue identifiers via outliers over MCP before the team-filtering rules are settled is too risky.

I'll happily park this PR. Once a decision is made on #317 and the underlying team filter is implemented, just ping me here. I'll update the adapter to respect the new security rules, add the three test cases you outlined in `packages/mcp-server/tests/`, and fix the registration ordering!

imshashank:
Decision made on #317, so this is unblocked in principle. Details are on the issue, but the short version:

Aggregates stay workspace-wide, the drilldown gets scoped to the reader's teams. Totals spanning the whole workspace is deliberate and stays; handing back identifiers and titles for teams you are not on does not.

What that means for this PR. The `metrics` half of your response maps to the aggregate path, so it is fine as written. The `outliers` half is drilldown-shaped, since it returns identifiers and titles, and that is the part that has to follow team membership once #317 lands.

I would not rebuild this on top of an unimplemented rule though, so ordering matters: #317 needs implementing first, then this adapter picks up the scoping for free if it goes through the same path. If you want to take #317 as well it is now labelled `ready` and `help wanted`, and you would be the natural person to do it given you have already read this code. Entirely your call, and no obligation.

One thing to carry across from #317 when you do come back to this: if a total says 14 and the drilldown returns 3 rows, the tool response needs to say so rather than quietly returning a short list. An agent reading a truncated list with no signal will state it as fact. Worth returning the withheld count alongside the rows.

The three tests and the registration ordering from my review still stand. Thanks for parking it cleanly rather than pushing back, that was the right read.

packages/mcp-server/src/tools/analytics.ts
```diff
@@ -0,0 +1,63 @@
...
+        lens: 'overview',
+        measure: args.measure,
+        range: { preset: args.range },
+        compare: 'auto',
+        filter: { kind: 'group', combinator: 'and', children: [] },
+        includeArchived: false,
+        includeCanceled: false,
+        focus: {},
+      });
+      const overview = await loadAnalyticsOverview(principal, query);
```
imshashank, line 40: This workspace-wide overview includes issue-level `outliers`, and the adapter returns their identifiers and titles below. I reproduced a guest on one team receiving another team's `OPS-1` and `Classified rollout` through this tool. Scope or omit issue-level rows here and expose a withheld count; #317's planned drilldown scoping alone does not cover this call path.

packages/mcp-server/src/tools/analytics.ts
```diff
@@ -0,0 +1,63 @@
...
+  defineTool(
+    server,
+    {
+      name: 'get_analytics_overview',
+      title: 'Get Analytics Overview',
+      description:
+        'Retrieve a high-level analytics overview of the workspace. Use this to answer questions about throughput, median cycle time, blocked work, and current WIP. You can filter the time range and choose to measure by issues or points.',
+      readOnly: true,
+      inputSchema: {
+        range: z
```
imshashank, line 18: These team-sprint presets cannot be resolved reliably because the tool accepts no team or cycle context. With multiple active sprints, core falls back to a 30-day range, while this response does not expose the resolved dates. Add a validated selector or remove these presets, and return the effective range.

imshashank (changes requested):
The strict input schema and `orbit.read` registration are sound, but this is not safe to merge yet.

I reproduced the permission issue through the actual MCP tool: a guest belonging only to the default team received another team's `OPS-1` identifier, title `Classified rollout`, and cycle time. #317 remains open, and changing only `listAnalyticsDrilldown` will not fix this adapter because it reads `loadAnalyticsOverview().outliers` directly.

The sprint presets are also ambiguous without a team or cycle selector, and the response hides the resolved date range. In a workspace with multiple active sprints, core intentionally falls back to 30 days, so an agent can report the wrong period without knowing it.

Please scope or omit issue-level outliers and include withheld/completeness metadata; add a validated team or cycle selector or remove the sprint presets; return stable metric IDs plus the resolved range; and add focused MCP coverage for scope registration, defaults/query translation, response shape, multi-team sprint behavior, and cross-team visibility. Please also merge current `main`, place the registration in the requested catalog order, rerun the incomplete CodeRabbit review, resolve the remaining thread, and get the complete check set green.

imshashank:
Status check, no action needed from me right now, this one's waiting on you.

My Aug 21 review is still current: the cross-team leak reproduces directly through this adapter (a guest in the default team received another team's `OPS-1` identifier, title, and cycle time), the sprint presets need a team/cycle selector or should be dropped since core's 30-day fallback is ambiguous across multiple active sprints, and the response should carry the resolved date range plus stable metric IDs. Flagging as `waiting-for-author` since there's been no update since that review.

imshashank:
Thanks for the update. The head is still `c398fce6`, so the existing changes-requested review remains current.

One dependency correction: #346 currently scopes only `listAnalyticsDrilldown`. This tool calls `loadAnalyticsOverview().outliers`, so #346 as written does not make those identifiers and titles safe. You can proceed independently by omitting raw outliers, or wait for a scoped overview contract with withheld metadata.

On the next push, please also resolve the sprint-preset ambiguity, return the effective range and stable metric IDs, add focused tests under `packages/mcp-server/tests/tools/analytics.test.ts`, merge current main, fix registration order, resolve all three threads, and rerun CodeRabbit plus full CI.

Please change `Closes #219` to `Addresses the analytics gap in #219`, since #219 tracks a broader unfinished audit.

SinghAstra:
i will take care of this by today

imshashank (changes requested):
Re-reviewed exact head `fa345a5b` against live `main` (`0f8e7094`). The raw outlier leak and ambiguous sprint presets are fixed, and the stable metric IDs are a good direction. I am retaining changes requested for the remaining response-contract and test gaps.

1. `resolvedRange` is not actually resolved. `packages/mcp-server/src/tools/analytics.ts:42-53` echoes `args.range`, so the default response reports `"auto"` rather than the effective start, end, and timezone. Core computes that range, but `AnalyticsOverview` does not carry it back to this tool.
2. The current CodeRabbit finding is valid: `packages/mcp-server/tests/tools/analytics.test.ts:48-64` checks only key presence. Stable IDs, the default `issues` unit, and `comparisonDelta` semantics could all regress while the test remains green.
3. Three of four review threads remain unresolved. The raw-outlier finding is fixed in code but still needs a reply and resolution; the range thread remains substantively open.
4. The branch is nine commits behind current `main` and must be updated before final review.

Please return the effective resolved range from core, strengthen the semantic assertions, respond to and resolve each thread, merge current `main`, and rerun every current-head gate before requesting re-review. Retaining `waiting-for-author`.

imshashank:
Checking in here. It's been a few days since "I will take care of this by today" on the 23rd with no update since. No rush if priorities shifted, just let me know either way so I know whether to keep this queued or park it.

Outstanding from my last review: the sprint-preset ambiguity needs a team/cycle selector (or drop the presets), the response should return the effective resolved range and stable metric IDs, and it needs focused tests under `packages/mcp-server/tests/tools/analytics.test.ts` for scope registration, defaults, and response shape. Leaving this as `waiting-for-author`.

---

packages/mcp-server/src/tools/analytics.ts
imshashank: Fixed at exact head `f8f542a9`. Sprint-only presets were removed from the MCP input schema. Core now returns the effective ISO `resolvedRange` and nullable prior-period `comparisonRange`, including timezone, and the MCP tests assert exact 90-day and comparison-boundary semantics.

packages/mcp-server/src/tools/analytics.ts
imshashank: Fixed at exact head `f8f542a9`. The MCP adapter now omits every raw issue-level outlier and returns a single `outliersWithheldCount` that combines permission-hidden rows with authorized rows intentionally omitted by MCP. The regression covers both visible and restricted outliers, and the full MCP suite passes (199 tests, 932 assertions).

packages/mcp-server/tests/tools/analytics.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const metrics = result['metrics'] as Record<string, unknown>[];
+
+    for (const card of metrics) {
+      expect(card).toHaveProperty('id');
+      expect(card).toHaveProperty('metric');
+      expect(card).toHaveProperty('value');
+      expect(card).toHaveProperty('unit');
+    }
+  });
```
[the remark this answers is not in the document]
imshashank: Fixed at exact head `f8f542a9`. The test now asserts the complete stable metric ID order, the default throughput label/value/`issues` unit, `comparisonDelta`, and points-mode value/unit/delta semantics. Focused MCP analytics tests pass with 60 assertions.

imshashank:
@greptileai review

imshashank:
Maintainer repair is now pushed at exact head `f8f542a9`. The effective-range contract, stable metric semantics, raw-outlier withholding, permission-hidden count, and current-main integration are covered and locally green. All inline threads have been answered and resolved. I am holding merge until exact-head Actions and both review bots complete.

imshashank:
@coderabbitai review

imshashank (approved):
Approved at exact head `f8f542a9` after a full maintainer review. The privacy leak, effective range contract, preset ambiguity, semantic coverage, and stale-main findings are fixed. All inline threads are resolved, Greptile reports 5/5 on this commit, CodeRabbit is green, and every hosted Actions job passed.

PR #356, opened 2026-08-24
feat(mcp): add workspace agent instructions

### LINKED ISSUE #218: Per workspace agent instructions, served over MCP
An agent connecting to Orbit over MCP gets sixty odd tools and no context. It does not know that this workspace puts a `product/` prefix on every label, that bugs go to the Platform team, that the estimate scale is Fibonacci, or that nobody is allowed to set P0 without saying why in the description.

So every agent working in an Orbit workspace rediscovers the workspace's conventions by trial and error, badly, every session.

## What to build

Per workspace instructions, served to any connected agent.

- A block of prose an admin writes, describing how this workspace works: naming, conventions, routing, what to do and what never to do
- Optionally per team, since teams differ
- Exposed to MCP clients through the mechanism the protocol provides for this. The MCP specification has instructions on the server and prompts as a primitive, and which one fits is a real question worth answering in the design comment rather than assuming
- Visible and editable in settings, so it is a document the team maintains rather than a hidden config

## Why this is worth doing

It is small, it is cheap, and it is the difference between an agent that files issues a team has to clean up and one that files issues a team keeps. Almost every complaint about AI agents in a tracker is really a complaint about missing context, and this is the cheapest possible fix for it.

It also composes with the issue templates work in the Planning parity milestone. A template is machine readable convention, instructions are human readable convention, and an agent wants both.

## Things to think about

- **Instructions are not permissions.** Text saying "do not close issues" is advice, not enforcement. Anything that must not happen is enforced in `packages/shared/src/policy` and by the OAuth scopes, and the instructions must not be presented as a security control
- **Size.** Instructions consume context on every agent session. Cap the length and say so in the editor
- **Untrusted content.** The instructions are written by workspace admins, so they are trusted relative to the workspace, but they are still injected into a model's context. Do not build anything that lets a lower privileged person edit them
- **Discovery.** An agent should be able to fetch this deliberately as well as receiving it automatically, because a long running client may connect once and work for hours

## Files

- `packages/mcp-server/src/server.ts`
- `packages/mcp-server/src/tools/identity.ts`
- `packages/db/src/schema/org.ts`
- `apps/web/src/app/(app)/settings/general/`
- `docs/mcp.md`

## Good scoped contribution

This is one of the smaller issues in the milestone and it does not depend on the agent identity work. A good way into the MCP package for somebody who has not touched it.

yxr-2025:
## What this changes

Adds workspace-scoped agent instructions for Orbit MCP, with administrator editing and read-only refresh support. @imshashank, thank you for the direction on keeping the first contribution limited to workspace scope.

## Why

Closes #218

Workspace guidance should be available to connected agents without sending Orbit data to an external service. This keeps the context in the workspace database, exposes it only through the selected workspace and granted MCP scope, and leaves authorization policy as the real security boundary.

## How you know it works

- Added organization.agentInstructions with an empty default and a 4,000 character limit.
- Added the General settings editor with live character count and administrator-only updates through org:manage.
- Added the production migration and an idempotent catch-up path that repairs a partially-created nullable column, backfills NULL values, restores the empty default, and enforces NOT NULL.
- Added dynamic MCP initialization instructions for new read-scoped connections.
- Added the read-only get_workspace_instructions tool, requiring orbit.read and workspace membership.
- Added workspace isolation, scope gating, discovery, refresh-after-update, four-thousand-character, membership, and adversarial authorization tests.
- Included organization update payload coverage and updated docs/mcp.md.

Targeted verification passed for MCP, MCP token authorization, organization services, validators, migration catch-up, and typechecking. `BETTER_AUTH_SECRET` and `ANALYTICS_CURSOR_SECRET` are now configured in the ignored local `.env`, and the full web suite passes with 2,286 tests and no failures.

`bun run verify` remains unchecked because it still exits with a nonzero status on native Windows. The confirmed remaining failures are existing cross-platform test assumptions outside Issue #218: two database tests expect POSIX `/catchup/...` path separators, and the production build guard exercises Unix shell syntax that Windows reports as `Redirection with no file`. This PR does not change those unrelated tests because doing so would expand the workspace-instructions contribution. The checkbox is intentionally left unchecked rather than reporting a full verification pass that did not occur.

Local database verification also passed against `localhost/orbit`. The existing development database first required the repository's idempotent `workspace-sprints-catchup.sql` historical reconciliation. After that, `bun run db:release` completed successfully, baselining the ledger with 14 migrations recorded and 0 pending migrations applied. `bun run db:check-drift` then confirmed that the local database contains every schema object and invariant required by this code. These results apply to the local development database only and do not claim production database verification.

## Screenshots

Not applicable. The settings change is covered by the existing settings UI and this PR does not require a visual review artifact.

## Checklist

- [x] No any, no non-null assertions
- [x] External input is parsed with a Zod schema from @orbit/shared
- [x] Authorization is enforced on the server through packages/shared/src/policy, not only in the UI
- [ ] bun run verify completed successfully on native Windows. It remains unchecked for the cross-platform failures explained above
- [x] bun run db:release and bun run db:check-drift passed against the local development database

This first contribution intentionally stays at workspace scope. Team-level overrides are out of scope and should be handled as a separate policy and context-budget decision.

<h3>Greptile Summary</h3>

Adds workspace-scoped agent instructions that administrators can edit and read-scoped MCP clients can retrieve.
- Persists validated instructions with migration and catch-up support.
- Adds optimistic concurrency control for administrator edits.
- Exposes instructions during MCP initialization and through a read-only refresh tool.
- Adds settings UI, authorization coverage, migration tests, and MCP documentation.

The PR appears safe to merge, with no blocking failure remaining.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| apps/web/src/features/settings/general-form.tsx | Adds the administrator editor, dirty-field tracking, refresh reconciliation, and version-aware submission while using the shared textarea primitive. |
| packages/core/src/org/organization-service.ts | Adds partial organization updates and conditional sync-ID conflict detection for instruction edits. |
| packages/mcp-server/src/server.ts | Adds workspace guidance to MCP initialization for authorized read-scoped connections. |
| packages/mcp-server/src/tools/identity.ts | Adds the read-only workspace-instructions retrieval tool within the existing caller and scope boundary. |
| packages/db/src/schema/org.ts | Adds the non-null workspace instruction field with an empty default. |
| packages/db/catchup/workspace-agent-instructions.sql | Provides idempotent repair, backfill, default restoration, and nullability enforcement for partially updated databases. |
| packages/db/drizzle/0013_strange_harrier.sql | Adds the production schema migration for workspace agent instructions. |

</details>

```mermaid
sequenceDiagram
  participant A as Workspace administrator
  participant W as Settings UI
  participant C as Organization service
  participant D as Database
  participant M as MCP client
  A->>W: Edit agent instructions
  W->>C: PATCH instructions and expected sync ID
  C->>D: Update when sync ID matches
  D-->>C: Updated instructions and new sync ID
  C-->>W: Updated workspace
  M->>C: Connect with workspace-bound read scope
  C->>D: Read workspace instructions
  D-->>C: Current instructions
  C-->>M: Initialization instructions
  M->>C: get_workspace_instructions
  C->>D: Refresh current instructions
  D-->>M: Workspace guidance
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(workspace): prevent stale instructio..."](https://github.com/noveum/orbit/commit/ca8e6395bdcd93c27a6ddd8afeccb335bd32d211) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56260554)</sub>

### DIFF
--- a/apps/web/src/features/settings/general-form.tsx
+++ b/apps/web/src/features/settings/general-form.tsx
@@ -2,10 +2,11 @@
 
 import { AGENT_INSTRUCTIONS_MAX_LENGTH } from '@orbit/shared/constants';
 import { useRouter } from 'next/navigation';
-import { type FormEvent, useState } from 'react';
+import { type FormEvent, useEffect, useRef, useState } from 'react';
 import { Badge } from '@/components/ui/badge.tsx';
 import { Button } from '@/components/ui/button.tsx';
 import { Input } from '@/components/ui/input.tsx';
+import { Textarea } from '@/components/ui/textarea.tsx';
 import { useToast } from '@/components/ui/toast.tsx';
 import { apiRequest, messageOf } from '@/lib/api/client.ts';
 
@@ -25,6 +26,7 @@ export interface GeneralFormProps {
   readonly logo: string | null;
   readonly allowedEmailDomains: readonly string[];
   readonly agentInstructions: string;
+  readonly syncId: number;
   readonly canManage: boolean;
 }
 
@@ -33,6 +35,7 @@ export function GeneralForm({
   logo,
   allowedEmailDomains,
   agentInstructions,
+  syncId,
   canManage,
 }: GeneralFormProps) {
   const router = useRouter();
@@ -44,6 +47,43 @@ export function GeneralForm({
   const [pending, setPending] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
+  const dirtyFields = useRef({
+    name: false,
+    logo: false,
+    allowedEmailDomains: false,
+    agentInstructions: false,
+  });
+  const nameBaseline = useRef(name);
+  const logoBaseline = useRef(logo ?? '');
+  const domainsBaseline = useRef(parseDomains(allowedEmailDomains.join(', ')));
+  const instructionsBaseline = useRef(agentInstructions);
+  const syncIdBaseline = useRef(syncId);
+
+  useEffect(() => {
+    nameBaseline.current = name;
+    if (!dirtyFields.current.name) setWorkspaceName(name);
+  }, [name]);
+
+  useEffect(() => {
+    logoBaseline.current = logo ?? '';
+    if (!dirtyFields.current.logo) setLogoUrl(logo ?? '');
+  }, [logo]);
+
+  useEffect(() => {
+    domainsBaseline.current = parseDomains(allowedEmailDomains.join(', '));
+    if (!dirtyFields.current.allowedEmailDomains) {
+      setDomains(allowedEmailDomains.join(', '));
+    }
+  }, [allowedEmailDomains]);
+
+  useEffect(() => {
+    if (!dirtyFields.current.agentInstructions) {
+      instructionsBaseline.current = agentInstructions;
+      syncIdBaseline.current = syncId;
+      setInstructions(agentInstructions);
+    }
+  }, [agentInstructions, syncId]);
+
   const parsedDomains = parseDomains(domains);
 
   async function onSubmit(event: FormEvent<HTMLFormElement>): Promise<void> {
@@ -51,15 +91,39 @@ export function GeneralForm({
     setPending(true);
     setError(null);
     try {
-      await apiRequest('/api/organizations/current', {
-        method: 'PATCH',
-        body: {
-          name: workspaceName,
-          logo: logoUrl.trim().length === 0 ? null : logoUrl.trim(),
-          allowedEmailDomains: parsedDomains,
-          agentInstructions: instructions,
+      const body: {
+        name?: string;
+        logo?: string | null;
+        allowedEmailDomains?: string[];
+        agentInstructions?: string;
+        expectedSyncId?: number;
+      } = {};
+      if (dirtyFields.current.name) body.name = workspaceName;
+      if (dirtyFields.current.logo) {
+        body.logo = logoUrl.trim().length === 0 ? null : logoUrl.trim();
+      }
+      if (dirtyFields.current.allowedEmailDomains) body.allowedEmailDomains = parsedDomains;
+      if (dirtyFields.current.agentInstructions) {
+        body.agentInstructions = instructions;
+        body.expectedSyncId = syncIdBaseline.current;
+      }
+
+      const result = await apiRequest<{ organization: { syncId: number } }>(
+        '/api/organizations/current',
+        {
+          method: 'PATCH',
+          body,
         },
-      });
+      );
+      nameBaseline.current = workspaceName;
+      logoBaseline.current = logoUrl;
+      domainsBaseline.current = parsedDomains;
+      instructionsBaseline.current = instructions;
+      syncIdBaseline.current = result.organization.syncId;
+      dirtyFields.current.name = false;
+      dirtyFields.current.logo = false;
+      dirtyFields.current.allowedEmailDomains = false;
+      dirtyFields.current.agentInstructions = false;
       toast({ title: 'Workspace updated', tone: 'success' });
       router.refresh();
     } catch (caught) {
@@ -84,7 +148,11 @@ export function GeneralForm({
           <Input
             id="settings-name"
             value={workspaceName}
-            onChange={(event) => setWorkspaceName(event.target.value)}
+            onChange={(event) => {
+              const value = event.target.value;
+              setWorkspaceName(value);
+              dirtyFields.current.name = value !== nameBaseline.current;
+            }}
             required
             minLength={2}
             maxLength={64}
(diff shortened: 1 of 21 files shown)


imshashank (commented):
Thanks for keeping the first contribution scoped to workspace level like we discussed on #218, and for the idempotent catch-up script with its own repair test, that is exactly the pattern the rest of `packages/db/catchup` uses.

I read through the full diff, not just the summary. The authorization is right where it matters: `packages/mcp-server/src/server.ts` gates initialization delivery on `grantsReads(scopes)`, `get_workspace_instructions` is registered `readOnly: true` through the existing tool-catalog gate, and `packages/mcp-server/tests/tools.test.ts` and `auth.test.ts` both prove a write-only token gets neither the initialization text nor the tool. The 4,000-character limit is enforced in both the Zod schema and the textarea's `maxLength`, and `apply-catchup.test.ts` covers the partial-column repair path. Core service tests correctly check both the admin-succeeds and member-forbidden cases.

Four things worth fixing, none of them blocking on their own, but I'd like them closed before merge since three are one-line fixes:

1. `general-form.tsx:148-150` renders the character counter with `.toLocaleString()` on both `instructions.length` and `AGENT_INSTRUCTIONS_MAX_LENGTH`. That's locale-dependent (thousands separators differ, e.g. `4,000` vs `4.000`), and since this is a client component the server-rendered and client-rendered locale can disagree, which is a hydration mismatch. Plain numeric text is fine here, nothing here needs grouping separators at this range.
2. `docs/mcp.md:64-67` says a newly connected client receives workspace instructions during initialization without qualifying that this only happens when the token grants `orbit.read`. The code is already correct, per point above; the sentence just needs the same qualifier the rest of that doc section uses.
3. No `apps/web/tests` coverage for the editor itself, rendering the field, editing it, and asserting the submitted payload carries `agentInstructions` including the boundary at 4,000 chars. Everything below the UI is well tested; this is the one layer that isn't.
4. Greptile flagged the textarea's transition classes being hand-rolled instead of going through `apps/web/src/lib/interaction.ts`. Worth checking whether the sibling inputs in this same form already do it this way, if so it's pre-existing and not this PR's problem to fix, but if this is a new textarea pattern, route it through the shared tokens per CLAUDE.md.

Labeling `waiting-for-author` for these four. Nothing else in this PR concerns me, once these land I'll do a final pass and this should be quick.

apps/web/src/features/settings/general-form.tsx
```diff
@@ -48,6 +57,7 @@ export function GeneralForm({ name, logo, allowedEmailDomains, canManage }: Gene
           name: workspaceName,
           logo: logoUrl.trim().length === 0 ? null : logoUrl.trim(),
           allowedEmailDomains: parsedDomains,
+          agentInstructions: instructions,
```
imshashank: [P1] Preserve concurrent instruction edits. This state is captured once, and every submission includes it even when the admin changed only the name, logo, or domains. Organization deltas do not refresh this value, so a stale tab can silently revert a newer admin edit. Reconcile it from a realtime current-organization cache or submit only dirty fields, and add a two-client regression.

packages/mcp-server/src/server.ts
```diff
@@ -40,10 +40,18 @@ export function grantsWrites(scopes: string): boolean {
   return granted(scopes).has(ORBIT_WRITE_SCOPE);
 }
 
-export function createOrbitMcpServer(principal: Principal, scopes = EVERY_ORBIT_SCOPE): McpServer {
+export async function createOrbitMcpServer(
+  principal: Principal,
+  scopes = EVERY_ORBIT_SCOPE,
+): Promise<McpServer> {
+  const organization = await getOrganization(principal.organizationId);
```
imshashank: [P2] Avoid loading the organization for every MCP request. Dispatch creates a fresh server for initialize, notifications, and tool calls, so this query runs even for write-only scopes where the value is discarded; the explicit refresh tool then queries again. Load instructions only for initialization or fold the field into an existing authenticated lookup without introducing stale caching.

imshashank (changes requested):
Current-head review of `b846e8b6`. The authorization, migration and catchup, MCP scope gating, and explicit refresh tool are sound. I am requesting changes for two additional blockers plus the four existing unresolved threads.

Please fix the stale settings overwrite and per-request organization query described inline. Then close the existing editor regression, locale-stable counter, shared Textarea and focus styling, and `orbit.read` documentation threads. Merge current `main`, move the head to a neutral branch name because `codex/...` violates this repository's no-AI-attribution rule, add both-theme visual evidence, run full verification, rerun CI and both review bots on the resulting exact head, and resolve every thread.

Current gates are not merge-ready: the branch is behind and diverged from `main`, four threads remain unresolved, CI, links, and CodeQL have not executed, Vercel requires authorization, and CodeRabbit is rate-limited rather than complete. Exact-head Greptile alone is insufficient. Do not merge this head.

PR #366, opened 2026-08-28
fix(auth): explain hosted access restriction (#344)

### LINKED ISSUE #344: [bug]: hosted sign-in hides the email-domain restriction behind a generic OAuth error
### What happened

I was trying to login to the website with google and github but after i select the accounts/authorize i am getting Sign in Failed toast

### What you expected instead

I expected properly sign in flow

### Steps to reproduce

1. try to sign in to orbit.noveum.ai
2. try to login with google/github
3. check the toast on the bottom left
4. you will get failed to login

### Where did you see this

orbit.noveum.ai

### Which part of Orbit

Authentication and accounts

### If this is a realtime problem

- [ ] A "Reconnecting to live updates" banner appeared
- [ ] More than one tab or browser was open
- [ ] More than one user was in the workspace

### Environment

```markdown
- Orbit commit or version:
- Bun version (bun --version):
- OS:
- Browser:
```

### Logs, errors or screenshots

```shell

```

### Before you submit

- [x] I searched the existing issues and this is not a duplicate
- [x] I can reproduce this consistently, or I have said how often it happens

mikemikimike:
## What this changes

Closes #344

- Show the hosted access restriction before sign-in on the landing and login surfaces.
- Keep the notice limited to the official hosted deployment so self-hosted instances are not mislabelled.
- Map the shared email-domain restriction code to an actionable message for Google and GitHub OAuth callbacks.
- Add regression coverage for both providers, first-time account rejection, existing-account session rejection, and the callback toast.

## Why

The hosted demo intentionally accepts only Noveum employee email domains, but the landing and login pages did not explain that policy. Ineligible accounts could therefore see a generic sign-in failure even when the OAuth provider was working.

## How know works

- `ORBIT_TEST_LANE=issue344final3 bun --env-file=../../.env test tests/app/page.test.tsx tests/components/auth/login-form.test.tsx tests/components/auth/auth-error-notice.test.tsx tests/lib/auth/oauth-error.test.ts tests/lib/auth/server.test.ts tests/lib/auth/server-session-domain.test.ts --timeout 20000` (40 passed)
- `bun run lint` (passed with existing warning/info output)
- `bun run typecheck` (passed)
- `bun run check-comments`, `bun run check-bytes`, `bun run check-bun-imports`, and `bun run check-deps` (passed)
- `bun --env-file=../../.env ../../packages/db/src/check-drift.ts --guard-deploy` (passed)
- `bun --env-file=../../.env next build` from `apps/web` (passed)
- `git diff --check` (passed)

## Screenshots

Not applicable; this is a small copy and behavior change with no layout or visual asset changes.

## Checklist

- [x] The server-side email-domain policy is unchanged.
- [x] The hosted notice is deployment-aware.
- [x] Regression tests cover the affected paths.

The repository's outer `bun run verify` and `bun run build` wrappers hit existing Windows POSIX-shell/test-environment limitations. The focused tests, static checks, typecheck, schema drift guard, and direct Next production build passed. Real Google/GitHub provider E2E was not run because it requires provider credentials and an external hosted account.

### DIFF
--- a/apps/web/src/app/(auth)/login/page.tsx
+++ b/apps/web/src/app/(auth)/login/page.tsx
@@ -5,9 +5,10 @@ import { DevSignIn } from '@/components/auth/dev-sign-in.tsx';
 import { LoginForm } from '@/components/auth/login-form.tsx';
 import { devLoginEnabled } from '@/lib/api/dev-login.ts';
 import { listDevUsers } from '@/lib/api/dev-users.ts';
-import { authErrorCode } from '@/lib/auth/oauth-error.ts';
+import { authErrorCode, hostedAccessNoticeFor } from '@/lib/auth/oauth-error.ts';
 import { enabledSocialProviders, passwordAuthEnabled } from '@/lib/auth/server.ts';
 import { getSession } from '@/lib/auth/session.ts';
+import { publicAppUrl } from '@/lib/env.ts';
 import { mcpContinueUrl, safeCallback } from './continue-url.ts';
 
 export const metadata: Metadata = { title: 'Sign in' };
@@ -24,12 +25,14 @@ export default async function LoginPage({
   if (session !== null) redirect(callbackUrl ?? '/my-issues');
 
   const devUsers = devLoginEnabled() ? await listDevUsers() : [];
+  const hostedAccessNotice = hostedAccessNoticeFor(publicAppUrl());
 
   return (
     <main className="flex min-h-dvh items-center justify-center bg-bg px-5 py-12">
       <div className="w-full max-w-sm rounded-xl border border-border bg-surface p-6 shadow-pop sm:p-7">
         <LoginForm
           providers={enabledSocialProviders}
+          {...(hostedAccessNotice === undefined ? {} : { hostedAccessNotice })}
           passwordEnabled={passwordAuthEnabled}
           {...(callbackUrl === undefined ? {} : { callbackUrl })}
         />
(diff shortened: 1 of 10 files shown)


imshashank:
I am thinking of changing this policy and allowing people to actually log in to Orbit.Noveum.ai.

imshashank:
Thank you for this, and apologies that it is landing this way.

The policy this PR explains is being removed rather than documented. #369 opens sign-up on orbit.noveum.ai to anyone, and `ALLOWED_EMAIL_DOMAINS` has already been removed from the production environment, so the notice added here would describe a restriction that no longer exists. That is a change of direction on our side, not a judgement on the work, which was accurate about the behaviour as it stood when you wrote it.

Two things from your branch that I do not want to lose.

Your reading of the failure was right, and more right than the issue gave you credit for. The generic result you were fixing was not only a copy problem. A sign-in code request from an address outside the allowlist was answering `200 {"success":true}` and sending no mail at all, because better-auth runs the email OTP callback through `runInBackgroundOrAwait`, which catches whatever the callback throws and only logs it. So the screen said a code was on its way and none ever arrived. #369 moves that check into a `before` hook where it returns a real `403`.

The `email_domain_not_allowed` message you rewrote is still worth having. The current copy says the domain is not allowed "on this workspace", but that error also fires for the server level allowlist, so the wording is wrong for exactly the case that produced this issue. That part is untouched by the policy change and still applies to any self-hosted instance that sets `ALLOWED_EMAIL_DOMAINS`. If you would like to send it as a small focused PR, I will review it quickly.

Closing this in favour of #369. Thanks again for the first contribution, and for writing tests with it.

PR #384, opened 2026-09-03
feat(mcp): add accurate tool annotations and update biome schema

ashutoshb206:
## What this changes

Adds explicit `destructive` and `idempotent` behavioral annotations to MCP tools across `workspace.ts`, `org.ts`, `docs.ts`, `issues.ts`, and `planning.ts` by extending `ToolConfig` in `packages/mcp-server/src/tools/support.ts` (`MCP-001`). Also updates `biome.json` schema version to 2.5.7 and cleans up test lint warnings.

## Why

Completes **MCP-001** from `docs/open-source-readiness.md` so connected MCP clients receive accurate safety hints (`destructiveHint`, `idempotentHint`) for destructive and idempotent write operations.

## How you know it works

- Ran `bun run check-comments` (0 disallowed comments).
- Ran `bun run check-bun-imports` (0 Bun built-ins in shipped server code).
- Ran `bun run check-deps` (6 overridden packages resolved).
- Ran `bun run typecheck` across all monorepo packages (`@orbit/shared`, `@orbit/db`, `@orbit/core`, `@orbit/services`, `@orbit/mcp-server`, `@orbit/web`).
- Ran `bun test packages/db/tests/check-source-bytes.test.ts` (9/9 tests passing).
- Ran `biome check .` (0 errors, 0 warnings).

## Checklist

- [x] `bun run verify` is green, all four checks
- [x] Tests added or updated, and they fail without the change
- [x] No comments added to code, and no em-dash characters anywhere
- [x] No `any`, no non-null assertions
- [x] External input is parsed with a Zod schema from `@orbit/shared`
- [x] Authorization is enforced on the server through `packages/shared/src/policy`, not only in the UI
- [x] Docs updated if behaviour, configuration or setup changed

## Anything reviewers should know

Write tools default `idempotentHint` to `readOnly || (idempotent ?? false)` and `destructiveHint` to `destructive ?? false`. Read tools automatically default `idempotentHint` to `true`.

<h3>Greptile Summary</h3>

The PR adds explicit MCP destructive-operation metadata, uses conservative defaults for unannotated write tools, and prevents duplicate relation activity on retries.
- Corrects destructive annotations for deletion and ordinary write tools.
- Removes inaccurate idempotency claims from document mutations.
- Updates annotation tests to use the registered `delete_label` and `delete_state` names.
- Updates the Biome schema and MCP readiness documentation.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

[...]

PR #440, opened 2026-09-07
feat(issues): support merging issues as duplicates

### LINKED ISSUE #418: Merge two issues as duplicates
Left over from #212 once #379 lands. Detection tells you two issues are the same; nothing lets you act on it in one move.

## What to build

- A "mark as duplicate of" action on an issue, from the detail view, the command palette and MCP.
- It records the `duplicate` relation in `issue_relation`, moves the duplicate to the team's canceled state with a `duplicate` reason, and re-points its subscribers at the survivor so they hear where it went.
- Comments stay on the duplicate; the survivor gets one activity row linking to it. Moving comments across is not worth the ambiguity.
- The duplicate's detail view shows a banner pointing at the survivor.

Blocked on #379. Check what `issue_relation` already supports for the duplicate type before adding anything.

Pallavikumarimdb:
## What this changes

Adds a **Mark duplicate** action to issues accessible from the detail view, command palette (`Shift+M`), and MCP (`mark_issue_duplicate`). 

Marking an issue as a duplicate creates reciprocal `duplicate_of` / `duplicated_by` entries in `issue_relation`, moves the duplicate issue to the team's canceled state (`Duplicate`), transfers issue subscribers to the survivor, appends relation activity logs, and displays a banner pointing to the survivor issue on the duplicate's detail view.

## Why

Detection previously indicated when issues were duplicates, but there was no single action to act on it in one step.

Closes #418

## How you know it works

- Added unit tests in `packages/core/tests/work/issue-duplicates.test.ts` covering subscriber transfer, status updates, activity creation, and duplicate relation creation.
- Added REST API route tests for `POST /api/issues/[id]/duplicate` in `apps/web/tests/app/api/issues/[id]/duplicate/route.test.ts`.
- Added MCP tool integration tests for `mark_issue_duplicate` in `packages/mcp-server/tests/tools.test.ts`.
- Manually verified UI workflows including the `Shift+M` hotkey, relation picker, and duplicate banner navigation.

## Screenshots
image
image
image

## Checklist

- Used existing `duplicate_of` and `duplicated_by` types from `@orbit/shared/constants` and `issue_relation` schema to prevent redundant relation definitions or database schema changes.
- Comments remain on the duplicate issue while one activity entry (`duplicated_by INS-...`) is appended to the survivor issue to avoid comment migration ambiguity.

<h3>Greptile Summary</h3>

The PR adds issue duplicate merging through the web interface, REST API, and MCP.
- Creates reciprocal duplicate relations and supports replacing a previous survivor.
- Moves the duplicate into a canceled workflow state and transfers its subscribers.
- Publishes relation, issue, and subscription changes for realtime clients.
- Displays duplicate status and navigation in issue details.
- Adds domain, route, and MCP integration coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/core/src/work/issue-service.ts | Implements transactional duplicate replacement, reciprocal relations, state transition, subscriber transfer, activity, and correctly scoped realtime actions. |
| apps/web/src/app/api/issues/[id]/duplicate/route.ts | Parses shared input validation inside the common error-handling boundary and publishes the resulting actions. |
| apps/web/src/lib/query/use-relations.ts | Adds the duplicate mutation with response validation and invalidation for both affected issues and relation queries. |
| apps/web/src/features/issues/issue-detail.tsx | Adds the permission-aware duplicate picker, keyboard shortcut, and duplicate banner to issue details. |
| packages/mcp-server/src/tools/issues.ts | Exposes duplicate merging through a write-capable MCP tool and publishes the domain deltas. |
| packages/shared/src/validators/issue.ts | Defines the shared survivor issue identifier schema used at external boundaries. |

</details>

```mermaid
sequenceDiagram
    participant User
    participant Entry as Web or MCP
    participant Service as Issue service
    participant DB as Database
    participant Realtime
    User->>Entry: Mark issue as duplicate
    Entry->>Service: markAsDuplicate(source, survivor)
    Service->>DB: Remove prior duplicate pair
    Service->>DB: Insert duplicate_of and duplicated_by
    Service->>DB: Move source to canceled state
    Service->>DB: Transfer subscriptions and append activity
    DB-->>Service: Updated issue and relations
    Service-->>Entry: Result and sync actions
    Entry->>Realtime: Publish sync actions
    Entry-->>User: Updated duplicate issue
```
</details>

<sub>Reviews (3): Last reviewed commit: ["fix(issues): improve duplicate flow vali..."](https://github.com/noveum/orbit/commit/712d90fccee79e7f731b32fa89c42d2dc4ca5ff2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61363472)</sub>

### DIFF
--- a/packages/core/src/work/issue-service.ts
+++ b/packages/core/src/work/issue-service.ts
@@ -27,6 +27,7 @@ import {
   issueBulkUpdateSchema,
   issueCreateSchema,
   issueFilterSchema,
+  issueMarkDuplicateSchema,
   issueMoveSchema,
   issueRelationSchema,
   issueSummaryQuerySchema,
@@ -2662,4 +2663,202 @@ export async function findDuplicateIssues(
     },
     similarity: Number(row.similarity),
   }));
+}
+
+export async function markAsDuplicate(
+  principal: Principal,
+  issueId: string,
+  input: unknown,
+): Promise<{ issue: IssueRow; relations: IssueRelationRow[]; actions: SyncAction[] }> {
+  assertCan(principal, 'issue:update');
+  const parsed = issueMarkDuplicateSchema.parse(input);
+  if (parsed.survivorIssueId === issueId) {
+    throw validationFailed('An issue cannot be marked as a duplicate of itself.');
+  }
+
+  return await db.transaction(async (tx) => {
+    const source = await loadIssue(tx, principal, issueId);
+    const target = await loadIssue(tx, principal, parsed.survivorIssueId);
+
+    const canceledStates = await tx
+      .select()
+      .from(schema.workflowState)
+      .where(
+        and(
+          eq(schema.workflowState.teamId, source.teamId),
+          eq(schema.workflowState.category, 'canceled'),
+        ),
+      )
+      .orderBy(asc(schema.workflowState.position));
+
+    const canceledState =
+      canceledStates.find((state) => state.name.toLowerCase() === 'duplicate') ?? canceledStates[0];
+
+    if (canceledState === undefined) {
+      throw validationFailed('The team has no canceled status.');
+    }
+
+    const syncId = await nextSyncId(tx);
+    const actor = await principalActor(tx, principal);
+    const now = new Date();
+
+    const relations = await tx
+      .insert(schema.issueRelation)
+      .values([
+        {
+          id: newId(),
+          organizationId: principal.organizationId,
+          issueId: source.id,
+          relatedIssueId: target.id,
+          type: 'duplicate_of',
+          syncId,
+        },
+        {
+          id: newId(),
+          organizationId: principal.organizationId,
+          issueId: target.id,
+          relatedIssueId: source.id,
+          type: 'duplicated_by',
+          syncId,
+        },
+      ])
+      .onConflictDoNothing()
+      .returning();
+
+    const stateChanged = source.stateId !== canceledState.id;
+    let updatedIssue = source;
+    let sourceState: { id: string; name: string } | undefined;
+    if (stateChanged) {
+      const [sourceStateRow] = await tx
+        .select({ id: schema.workflowState.id, name: schema.workflowState.name })
+        .from(schema.workflowState)
+        .where(eq(schema.workflowState.id, source.stateId))
+        .limit(1);
+      sourceState = sourceStateRow;
+
+      const [updatedRow] = await tx
+        .update(schema.issue)
+        .set({
+          stateId: canceledState.id,
+          ...stateTimestamps('canceled', now),
+          updatedAt: now,
+          syncId,
+        })
+        .where(
+          and(
+            eq(schema.issue.id, source.id),
+            eq(schema.issue.organizationId, principal.organizationId),
+          ),
+        )
+        .returning();
+      if (updatedRow !== undefined) {
+        updatedIssue = updatedRow;
+      }
+    }
+
+    const existingSubs = await tx
+      .select({ userId: schema.issueSubscription.userId })
+      .from(schema.issueSubscription)
+      .where(eq(schema.issueSubscription.issueId, source.id));
+
+    const subUserIds = existingSubs.map((sub) => sub.userId);
+    const subActions: SyncAction[] = [];
+
+    if (subUserIds.length > 0) {
+      await subscribeUsers(tx, target.id, subUserIds, syncId);
+      await tx
+        .delete(schema.issueSubscription)
+        .where(eq(schema.issueSubscription.issueId, source.id));
+
+      for (const userId of subUserIds) {
+        subActions.push(
+          buildSyncAction({
+            syncId,
+            organizationId: principal.organizationId,
+            scopes: [scopes.user(userId)],
+            action: 'delete',
+            model: 'issue_subscription',
+            modelId: `${source.id}:${userId}`,
+            data: { issueId: source.id, userId, syncId },
+            actor,
+          }),
+          buildSyncAction({
+            syncId,
+            organizationId: principal.organizationId,
+            scopes: [scopes.user(userId)],
+            action: 'insert',
+            model: 'issue_subscription',
+            modelId: `${target.id}:${userId}`,
+            data: { issueId: target.id, userId, identifier: target.identifier, syncId },
+            actor,
+          }),
+        );
+      }
+    }
+
+    const activitiesToAppend: Parameters<typeof appendActivities>[1] = [
+      {
+        organizationId: principal.organizationId,
+        issueId: source.id,
+        actor,
+        field: 'relation',
+        from: null,
+        to: `duplicate_of ${target.identifier}`,
+        syncId,
+      },
+      {
+        organizationId: principal.organizationId,
+        issueId: target.id,
+        actor,
+        field: 'relation',
+        from: null,
+        to: `duplicated_by ${source.identifier}`,
+        syncId,
+      },
+    ];
+
+    if (stateChanged) {
+      activitiesToAppend.push({
+        organizationId: principal.organizationId,
+        issueId: source.id,
+        actor,
+        field: 'stateId',
+        from: sourceState ?? source.stateId,
+        to: { id: canceledState.id, name: canceledState.name },
+        syncId,
+      });
+    }
+
+    await appendActivities(tx, activitiesToAppend);
+
+    const relationActions = relations.map((row) =>
+      buildSyncAction({
+        syncId,
+        organizationId: principal.organizationId,
+        scopes: relationScopes(row, source, target),
+        action: 'insert',
+        model: 'issue_relation',
+        modelId: row.id,
+        data: row,
+        actor,
+      }),
+    );
+
+    const issueAction = buildSyncAction({
+      syncId,
+      organizationId: principal.organizationId,
+      scopes: [scopes.team(updatedIssue.teamId), scopes.issue(updatedIssue.id)],
+      action: 'update',
+      model: 'issue',
+      modelId: updatedIssue.id,
+      data: updatedIssue,
+      actor,
+    });
+
+    return {
+      issue: updatedIssue,
+      relations,
+      actions: [issueAction, ...relationActions, ...subActions],
+    };
+  });
 }
--- a/packages/core/src/work/issue-service.ts
+++ b/packages/core/src/work/issue-service.ts
@@ -2665,6 +2665,101 @@ export async function findDuplicateIssues(
   }));
 }
 
+async function removeExistingDuplicates(
+  tx: Executor,
+  organizationId: string,
+  source: IssueRow,
+  actor: Actor,
+  syncId: number,
+): Promise<SyncAction[]> {
+  const existingDuplicates = await tx
+    .select()
+    .from(schema.issueRelation)
+    .where(
+      and(
+        eq(schema.issueRelation.organizationId, organizationId),
+        or(
+          and(
+            eq(schema.issueRelation.issueId, source.id),
+            eq(schema.issueRelation.type, 'duplicate_of'),
+          ),
+          and(
+            eq(schema.issueRelation.relatedIssueId, source.id),
+            eq(schema.issueRelation.type, 'duplicated_by'),
+          ),
+        ),
+      ),
+    );
+
+  if (existingDuplicates.length === 0) return [];
+
+  await tx.delete(schema.issueRelation).where(
+    and(
+      eq(schema.issueRelation.organizationId, organizationId),
+      inArray(
+        schema.issueRelation.id,
+        existingDuplicates.map((row) => row.id),
+      ),
+    ),
+  );
+
+  return existingDuplicates.map((row) =>
+    buildSyncAction({
+      syncId,
+      organizationId,
+      scopes: [scopes.team(source.teamId), scopes.issue(source.id)],
+      action: 'delete',
+      model: 'issue_relation',
+      modelId: row.id,
+      data: { id: row.id, issueId: row.issueId, relatedIssueId: row.relatedIssueId },
+      actor,
+    }),
+  );
+}
+
+async function transferSubscriptions(
+  tx: Executor,
+  organizationId: string,
+  source: IssueRow,
+  target: IssueRow,
+  actor: Actor,
+  syncId: number,
+): Promise<SyncAction[]> {
+  const existingSubs = await tx
+    .select({ userId: schema.issueSubscription.userId })
+    .from(schema.issueSubscription)
+    .where(eq(schema.issueSubscription.issueId, source.id));
+
+  const subUserIds = existingSubs.map((sub) => sub.userId);
+  if (subUserIds.length === 0) return [];
+
+  await subscribeUsers(tx, target.id, subUserIds, syncId);
+  await tx.delete(schema.issueSubscription).where(eq(schema.issueSubscription.issueId, source.id));
+
+  return subUserIds.flatMap((userId) => [
+    buildSyncAction({
+      syncId,
+      organizationId,
+      scopes: [scopes.user(userId)],
+      action: 'delete',
+      model: 'issue_subscription',
+      modelId: `${source.id}:${userId}`,
+      data: { issueId: source.id, userId, syncId },
+      actor,
+    }),
+    buildSyncAction({
+      syncId,
+      organizationId,
+      scopes: [scopes.user(userId)],
+      action: 'insert',
+      model: 'issue_subscription',
+      modelId: `${target.id}:${userId}`,
+      data: { issueId: target.id, userId, identifier: target.identifier, syncId },
+      actor,
+    }),
+  ]);
+}
+
 export async function markAsDuplicate(
   principal: Principal,
   issueId: string,
@@ -2701,6 +2796,14 @@ export async function markAsDuplicate(
     const syncId = await nextSyncId(tx);
     const actor = await principalActor(tx, principal);
     const now = new Date();
+
+    const oldRelationDeleteActions = await removeExistingDuplicates(
+      tx,
+      principal.organizationId,
+      source,
+      actor,
+      syncId,
+    );
 
     const relations = await tx
       .insert(schema.issueRelation)
@@ -2756,45 +2859,14 @@ export async function markAsDuplicate(
       }
     }
 
-    const existingSubs = await tx
-      .select({ userId: schema.issueSubscription.userId })
-      .from(schema.issueSubscription)
-      .where(eq(schema.issueSubscription.issueId, source.id));
-
-    const subUserIds = existingSubs.map((sub) => sub.userId);
-    const subActions: SyncAction[] = [];
-
-    if (subUserIds.length > 0) {
-      await subscribeUsers(tx, target.id, subUserIds, syncId);
-      await tx
-        .delete(schema.issueSubscription)
-        .where(eq(schema.issueSubscription.issueId, source.id));
-
-      for (const userId of subUserIds) {
-        subActions.push(
-          buildSyncAction({
-            syncId,
-            organizationId: principal.organizationId,
-            scopes: [scopes.user(userId)],
-            action: 'delete',
-            model: 'issue_subscription',
-            modelId: `${source.id}:${userId}`,
-            data: { issueId: source.id, userId, syncId },
-            actor,
-          }),
-          buildSyncAction({
-            syncId,
-            organizationId: principal.organizationId,
-            scopes: [scopes.user(userId)],
-            action: 'insert',
-            model: 'issue_subscription',
-            modelId: `${target.id}:${userId}`,
-            data: { issueId: target.id, userId, identifier: target.identifier, syncId },
-            actor,
-          }),
-        );
-      }
-    }
+    const subActions = await transferSubscriptions(
+      tx,
+      principal.organizationId,
+      source,
+      target,
+      actor,
+      syncId,
+    );
 
     const activitiesToAppend: Parameters<typeof appendActivities>[1] = [
       {
@@ -2858,7 +2930,7 @@ export async function markAsDuplicate(
     return {
       issue: updatedIssue,
       relations,
-      actions: [issueAction, ...relationActions, ...subActions],
+      actions: [issueAction, ...relationActions, ...subActions, ...oldRelationDeleteActions],
     };
   });
 }
--- a/packages/core/src/work/issue-service.ts
+++ b/packages/core/src/work/issue-service.ts
@@ -2802,11 +2802,53 @@ async function assertSurvivorAllowed(
     if (row === undefined) {
       cursor = null;
     } else {
+      if (row.relatedIssueId === sourceId) {
+        throw validationFailed(
+          'An issue cannot be marked as a duplicate of an issue that duplicates it.',
+        );
+      }
       if (cursor === target.id) {
         throw validationFailed('A duplicate issue cannot be a survivor.');
       }
       cursor = row.relatedIssueId;
     }
+  }
+}
+
+async function loadIssuesForUpdate(
+  tx: Executor,
+  principal: Principal,
+  sourceId: string,
+  targetId: string,
+): Promise<{ source: IssueRow; target: IssueRow }> {
+  const [firstId, secondId] = sourceId < targetId ? [sourceId, targetId] : [targetId, sourceId];
+  const firstIssue = await loadIssueForUpdate(tx, principal, firstId);
+  const secondIssue = await loadIssueForUpdate(tx, principal, secondId);
+  return {
+    source: sourceId === firstId ? firstIssue : secondIssue,
+    target: targetId === firstId ? firstIssue : secondIssue,
+  };
+}
+
+async function assertSourceAllowed(
+  tx: Executor,
+  organizationId: string,
+  sourceId: string,
+): Promise<void> {
+  const [sourceHasDuplicates] = await tx
+    .select()
+    .from(schema.issueRelation)
+    .where(
+      and(
+        eq(schema.issueRelation.organizationId, organizationId),
+        eq(schema.issueRelation.issueId, sourceId),
+        eq(schema.issueRelation.type, 'duplicated_by'),
+      ),
+    )
+    .limit(1);
+
+  if (sourceHasDuplicates !== undefined) {
+    throw validationFailed('An issue with duplicates cannot be marked as a duplicate.');
   }
 }
 
@@ -2822,8 +2864,12 @@ export async function markAsDuplicate(
   }
 
   return await db.transaction(async (tx) => {
-    const source = await loadIssueForUpdate(tx, principal, issueId);
-    const target = await loadIssue(tx, principal, parsed.survivorIssueId);
+    const { source, target } = await loadIssuesForUpdate(
+      tx,
+      principal,
+      issueId,
+      parsed.survivorIssueId,
+    );
 
     const [existingDuplicateOf] = await tx
       .select()
@@ -2842,6 +2888,7 @@ export async function markAsDuplicate(
     }
 
     await assertSurvivorAllowed(tx, principal.organizationId, source.id, target);
+    await assertSourceAllowed(tx, principal.organizationId, source.id);
 
     const canceledStates = await tx
       .select()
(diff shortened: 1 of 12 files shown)


packages/core/tests/work/issue-duplicates.test.ts
```diff
@@ -104,3 +113,55 @@ describe('findDuplicateIssues', () => {
...
+    const survRelations = await listRelatedIssues(workspace.admin, survivor.id);
+    expect(survRelations).toHaveLength(1);
+    expect(survRelations[0]?.type).toBe('duplicated_by');
+    expect(survRelations[0]?.issue.id).toBe(duplicate.id);
+
+    const survivorSubs = await listSubscribers(workspace.admin, survivor.id);
+    expect(survivorSubs.map((s) => s.userId)).toContain(workspace.admin.userId);
+
+    const dupSubs = await listSubscribers(workspace.admin, duplicate.id);
+    expect(dupSubs).toHaveLength(0);
```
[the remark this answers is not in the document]
Pallavikumarimdb: Addressed in [712d90f](https://github.com/Noveum/orbit/pull/440/commits/712d90fccee79e7f731b32fa89c42d2dc4ca5ff2)

packages/core/tests/work/issue-duplicates.test.ts
```diff
@@ -104,3 +113,55 @@ describe('findDuplicateIssues', () => {
...
+    const { issue: duplicate } = await createIssue(workspace.admin, {
+      teamId: workspace.teamId,
+      title: 'Duplicate Bug Report',
+    });
+
+    const result = await markAsDuplicate(workspace.admin, duplicate.id, {
+      survivorIssueId: survivor.id,
+    });
+
+    expect(result.issue.id).toBe(duplicate.id);
```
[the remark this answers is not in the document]
Pallavikumarimdb: Addressed in [712d90f](https://github.com/Noveum/orbit/pull/440/commits/712d90fccee79e7f731b32fa89c42d2dc4ca5ff2)

packages/core/src/work/issue-service.ts
```diff
@@ -2631,3 +2632,201 @@ export async function findDuplicateIssues(
...
+        {
+          id: newId(),
+          organizationId: principal.organizationId,
+          issueId: target.id,
+          relatedIssueId: source.id,
+          type: 'duplicated_by',
+          syncId,
+        },
+      ])
+      .onConflictDoNothing()
```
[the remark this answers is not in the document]
Pallavikumarimdb: Addressed in [712d90f](https://github.com/Noveum/orbit/pull/440/commits/712d90fccee79e7f731b32fa89c42d2dc4ca5ff2)

imshashank:
Status check: caught up on this. All nine review threads are resolved, Greptile's pass is 5/5 with "no blocking failure remains," and going through the diff against 712d90f, the fixes match what was asked for: the duplicate flow now reconciles an existing `duplicate_of` relation before inserting a new one instead of leaving two survivors linked, the route parses the body with `issueMarkDuplicateSchema` inside `handle` so validation errors map to the shared 422 response, the picker and `Shift+M` hotkey are gated behind `issue:update`, and the added test coverage (state transition, `canceledAt`, and the duplicate-only subscriber case) looks right.

Closes #418, so that's already linked.

The one thing I can't move from here: only Greptile Review and the label/reconcile jobs have run on this head so far. That's the same first-time-contributor gate #384 hit, `CI`, `CodeQL`, `Check links`, and `Publish documentation` are sitting in `action_required` until a maintainer approves the workflow runs from the Actions tab, and I don't have that permission through this integration. Marking this `needs-review` since it's on us now, not on you. Thanks for turning the feedback around fast, @Pallavikumarimdb.

---

imshashank:
Status check: nothing has moved here since my last note, still gated on the same first-time-contributor `action_required` approval for `CI`, `CodeQL`, `Check links` and `Publish documentation` on `712d90f`, which I can't grant through this integration. Keeping this at `needs-review` and flagging again for a maintainer to approve the workflow runs from the Actions tab so this can move.

---

imshashank (changes requested):
Thanks for working through every Greptile and CodeRabbit thread, they are all genuinely addressed. I read `markAsDuplicate` end to end against the rest of `issue-service.ts` and there is one blocking bug plus a few places where this mutation skips what every comparable mutation in the file does.

Blocking:
1. Duplicate cycles are not prevented (`packages/core/src/work/issue-service.ts`, `markAsDuplicate`). With A already `duplicate_of` B, calling `markAsDuplicate(B, { survivorIssueId: A })` succeeds: `existingDuplicateOf` only looks for `(issueId = B, type = duplicate_of)`, and `removeExistingDuplicates` never matches the `(B, A, duplicated_by)` row. Result: four relation rows, both issues canceled, both banners pointing at each other. Longer chains (A to B to C to A) have the same hole. Please refuse a survivor that is itself `duplicate_of` the source, directly or transitively (the sub-issue code in this file already walks parents the same way), and add a test for A to B then B to A.

Should fix in this PR (small, and they keep this consistent with `moveIssue` and `updateIssue`):
2. Validate the survivor: refuse an archived issue and an issue that is itself a duplicate (or re-point to the end of the chain).
3. Use `loadIssueForUpdate` for the source. This mutation changes the issue row, deletes relations and moves subscriptions, so it needs the row lock the other mutations take.
4. Emit the status-change notification through the existing `issueNotifications` path. Every other move to a canceled state notifies subscribers; this one is silent.
5. Build the issue update action with `issueAction` so it carries `labelIds` and `reviewerIds`. The client merge in `apps/web/src/lib/query/sync.ts` drops an inserted row that lacks them, so a view that did not already contain the issue never receives it.

Nits, take or leave:
- Move the same-survivor early return above `nextSyncId` so a no-op does not consume a sync id.
- `issue-detail.tsx` and `issue-relations.tsx` each mount their own picker and `useMarkDuplicate`; the hotkey could open the Links picker instead.
- Tests worth adding: forbidden role on the route and the MCP tool, cross-team re-point sync scopes, and the canceled state via the MCP tool.

Process: CI, CodeQL and the other required workflows have never run on this head because of the first-contribution gate, and CodeRabbit's last pass was rate limited. We will approve the runs on our side; please push the fixes above and we will re-review on the new head.

---

Pallavikumarimdb:
> Thanks for working through every Greptile and CodeRabbit thread, they are all genuinely addressed. I read `markAsDuplicate` end to end against the rest of `issue-service.ts` and there is one blocking bug plus a few places where this mutation skips what every comparable mutation in the file does.
> 
> Blocking:
> 
>     1. Duplicate cycles are not prevented (`packages/core/src/work/issue-service.ts`, `markAsDuplicate`). With A already `duplicate_of` B, calling `markAsDuplicate(B, { survivorIssueId: A })` succeeds: `existingDuplicateOf` only looks for `(issueId = B, type = duplicate_of)`, and `removeExistingDuplicates` never matches the `(B, A, duplicated_by)` row. Result: four relation rows, both issues canceled, both banners pointing at each other. Longer chains (A to B to C to A) have the same hole. Please refuse a survivor that is itself `duplicate_of` the source, directly or transitively (the sub-issue code in this file already walks parents the same way), and add a test for A to B then B to A.
> 
> 
> Should fix in this PR (small, and they keep this consistent with `moveIssue` and `updateIssue`): 2. Validate the survivor: refuse an archived issue and an issue that is itself a duplicate (or re-point to the end of the chain). 3. Use `loadIssueForUpdate` for the source. This mutation changes the issue row, deletes relations and moves subscriptions, so it needs the row lock the other mutations take. 4. Emit the status-change notification through the existing `issueNotifications` path. Every other move to a canceled state notifies subscribers; this one is silent. 5. Build the issue update action with `issueAction` so it carries `labelIds` and `reviewerIds`. The client merge in `apps/web/src/lib/query/sync.ts` drops an inserted row that lacks them, so a view that did not already contain the issue never receives it.
> 
> Nits, take or leave:
> 
>     * Move the same-survivor early return above `nextSyncId` so a no-op does not consume a sync id.
> 
>     * `issue-detail.tsx` and `issue-relations.tsx` each mount their own picker and `useMarkDuplicate`; the hotkey could open the Links picker instead.
> 
>     * Tests worth adding: forbidden role on the route and the MCP tool, cross-team re-point sync scopes, and the canceled state via the MCP tool.
> 
> 
> Process: CI, CodeQL and the other required workflows have never run on this head because of the first-contribution gate, and CodeRabbit's last pass was rate limited. We will approve the runs on our side; please push the fixes above and we will re-review on the new head.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

@imshashank Addressed the changes. Please review again

imshashank (approved):
All five points from my last review are in, and I read `markAsDuplicate` end to end against `6ce15a1c` to confirm rather than take the diff at its word:

1. Cycle prevention: `assertSurvivorAllowed` walks the `duplicate_of` chain from the target upward with a `visited` set, and rejects both the direct case and the transitive one. Test `'rejects creating duplicate cycles (A to B then B to A)'` covers it.
2. Survivor validation: archived and already-duplicate survivors are both refused, with a matching test each.
3. Row lock: `loadIssuesForUpdate` now locks source and target in canonical id order through `loadIssueForUpdate`, so this is consistent with `moveIssue`/`updateIssue` and safe against the cross-call deadlock a fixed lock order would otherwise risk.
4. Notification: `issueNotifications` runs right after the state update and before `transferSubscriptions` deletes the source's subscription rows, so the notify path still sees the original subscribers.
5. Sync action: `issueAction` now carries `labelIds`/`reviewerIds` from `issueDecorationsByIssue`, so a view that doesn't already hold the issue will receive it.

One more validation landed that I hadn't asked for but is the right call: `assertSourceAllowed` refuses marking an issue as a duplicate if it already has `duplicated_by` relations, closing the "former survivor becomes a duplicate" gap Greptile flagged separately.

CI is green across all 13 checks on this head, and all 13 bot review threads are resolved. Approving.

One process note: Greptile's trial expired partway through this PR, so its last few passes are "trial ended" rather than actual re-analysis. Nothing here suggests that matters, CodeRabbit and my own read both come back clean, but flagging it since it means Greptile hasn't actually seen the last two pushes.

---

packages/core/src/work/issue-service.ts
```diff
@@ -2631,3 +2632,400 @@ export async function findDuplicateIssues(
...
+  target: IssueRow,
+  actor: Actor,
+  syncId: number,
+): Promise<SyncAction[]> {
+  const existingSubs = await tx
+    .select({ userId: schema.issueSubscription.userId })
+    .from(schema.issueSubscription)
+    .where(eq(schema.issueSubscription.issueId, source.id));
+
+  const subUserIds = existingSubs.map((sub) => sub.userId);
```
imshashank: [P2] Preserve transferred subscribers when replacing the survivor

The first markAsDuplicate(A, B) transfers A subscribers to B and deletes every A subscription. A later supported markAsDuplicate(A, C) reads the now-empty A subscriptions and returns here, so those users remain on B and never follow C even though A points to C. I reproduced this using a member subscribed only to A; the replacement survivor lacks that member. Preserve enough transfer provenance to move the relevant subscriptions on replacement, or explicitly reject replacement after transfer if that is the intended contract. Extend the existing replacement-survivor test to assert subscriber behavior, without moving unrelated subscribers of B.

imshashank (changes requested):
Reviewed head 6ce15a1c across duplicate validation, row locking, reciprocal relations, notifications, subscriptions and REST/MCP entry points. All 12 core duplicate tests and 2 route tests pass locally. The initial transfer is covered well, but a temporary regression shows a duplicate-only subscriber does not follow a replacement survivor. See inline finding.

Bring in current main (2 commits behind), add the replacement-subscriber regression and rerun checks. The latest Greptile attempts report an expired trial rather than a completed review. Restore that review gate before merging; a subscription notice is not a code review under CLAUDE.md.

packages/core/src/work/issue-service.ts
Pallavikumarimdb: [8c2f081](https://github.com/Noveum/orbit/pull/440/commits/8c2f081045befc22f87285d05c718722a9f8006b)

Pallavikumarimdb:
@imshashank Addressed the changes. Please review again

packages/core/src/work/issue-service.ts
```diff
@@ -2661,3 +2662,500 @@ export async function findDuplicateIssues(
...
+          modelId: `${target.id}:${userId}`,
+          data: { issueId: target.id, userId, identifier: target.identifier, syncId },
+          actor,
+        }),
+      ),
+    );
+
+    await appendActivities(
+      tx,
+      allowedUserIds.map((userId) => ({
```
imshashank, line 2910: [P2] Record actual transfer ownership rather than every candidate subscriber. If user U already subscribes independently to A and B, marking A duplicate of B leaves the existing B row in place via onConflictDoNothing, but this records U as transferred anyway. Repointing A to C later deletes the independent B subscription at lines 2843-2849. Additionally, reading all historical transfers at lines 2747-2759 can re-subscribe a user who explicitly unsubscribed after the first transfer. Track current transfer provenance separately from independent subscriptions and preserve explicit opt-outs; cover both cases.

imshashank (changes requested):
Re-reviewed the latest subscription-transfer changes. The original A-to-B-to-C example is covered now, but the provenance still conflates independent subscriptions with transferred subscriptions. Please fix the inline case and add overlapping-subscriber and explicit-unsubscribe regression tests. We can run final validation after the author pushes the fix.

packages/core/src/work/issue-service.ts
Pallavikumarimdb: Updated in [0c92c0d](https://github.com/Noveum/orbit/pull/440/commits/0c92c0db98dbde4d02622c53e82c933435de8d0c) :

1. **Transfer Provenance**: We now distinguish pre-existing independent target subscriptions during transfer (`from: ${source.id}:independent`). Cleanup on repointing only deletes subscriptions created by the transfer (`fromValue === source.id`), preserving independent subscriptions on the previous survivor.
2. **Explicit Opt-Outs**: Candidate subscribers from previous survivors are now checked against active subscriptions on that survivor (`prevActiveSubs`), preventing users who explicitly unsubscribed from being re-subscribed when repointing.

Added regression tests covering both overlapping subscribers and explicit unsubscribes in `issue-duplicates.test.ts`. All 14 duplicate tests and `bun run verify` checks pass.

packages/core/src/work/issue-service.ts
```diff
@@ -2663,3 +2664,584 @@ export async function findDuplicateIssues(
...
+        organizationId,
+        issueId: source.id,
+        actor: { type: 'user', id: userId, name: actor.name },
```
imshashank: [P2] These provenance rows land in the user-facing activity log. `describeActivity` has no renderer for `subscription_transfer`, so the duplicate's timeline shows one line per subscriber, like `Ada Admin changed subscription_transfer from 4ee23354-…:independent to 18f97844-…`. `actorId` is the subscriber while `actorName` is whoever marked the duplicate, so each row credits the subscriber with something they did not do. The aiOnly filter in `packages/core/src/work/issue-query.ts:97` keys on `actorId`, so an agent member that only subscribed to A makes A show up as AI-touched once a human marks it. The creator is auto-subscribed, so every mark writes these rows. Move the provenance off `issue_activity` (see the summary), or at minimum exclude this field from `listActivityPage` and from the aiOnly predicate, with a test for each.

packages/core/src/work/issue-service.ts
```diff
@@ -2663,3 +2664,584 @@ export async function findDuplicateIssues(
...
+        field: 'subscription_transfer',
+        from: existingTargetUserIds.has(userId) ? `${source.id}:independent` : source.id,
```
imshashank: [P2] When two duplicates are merged into the same survivor, this labels the second transfer `:independent` because the user is already on B, even though the first transfer is what put them there. Repro: U follows A1 and A2, mark A1 as a duplicate of B, mark A2 as a duplicate of B, then re-point A1 to C. U is removed from B while A2 is still a duplicate of B, and U's subscription on A2 was already deleted by the second transfer, so U no longer follows the A2 thread anywhere. Re-pointing A2 to D later cannot bring it back, because `getPreviousSurvivorSubscribers` needs a live row on B.

packages/core/src/work/issue-service.ts
```diff
@@ -2663,3 +2664,584 @@ export async function findDuplicateIssues(
...
+
+    const prevTransferActivities = await tx
+      .select({ actorId: schema.issueActivity.actorId })
+      .from(schema.issueActivity)
+      .where(
+        and(
+          eq(schema.issueActivity.organizationId, organizationId),
+          eq(schema.issueActivity.issueId, sourceId),
+          eq(schema.issueActivity.field, 'subscription_transfer'),
+          eq(schema.issueActivity.fromValue, sourceId),
```
imshashank: [P2] This still treats any old `from = source` row plus a current row on the survivor as "held by the transfer". Nothing ever supersedes a transfer row, because `subscribe`, `unsubscribe`, assignment and comment auto-subscribe write no activity. Reproduced on this head:
- Mark A as a duplicate of B, U unsubscribes from B and then subscribes again, re-point A to C: U's B subscription is deleted.
- A to B to C, U unsubscribes from C, A to B, U subscribes to B, A to C: U is put back on C, which they left, and loses B, which they chose.
- D to A, D to B, U subscribes to A, D to A again (recorded as `D:independent`), D to C: U is removed from A because the first D to A row still matches.

So the explicit opt-out case from my last review is only covered when the user never subscribes again. Both `issue_subscription` and `issue_activity` carry `syncId`, so comparing the subscription's `syncId` with the latest matching transfer is one way to get this right without a schema change, if you keep the current design.

packages/core/src/work/issue-service.ts
```diff
@@ -2663,3 +2664,584 @@ export async function findDuplicateIssues(
...
+        from: existingTargetUserIds.has(userId) ? `${source.id}:independent` : source.id,
+        to: target.id,
+        syncId,
+      })),
+    );
+  }
+
+  if (existingSubs.length > 0) {
+    await tx
+      .delete(schema.issueSubscription)
```
imshashank: [P2] This deletes every subscription on the duplicate, including the users filtered out at line 2880 because they cannot read the survivor. They end up following neither issue while they can still read A. The same happens on a re-point: `cleanupPreviousSurvivorSubscriptions` removes a user from B even when the new survivor C is in a team they cannot read, and no transfer row to C is written, so a later re-point cannot bring them back. A support ticket marked as a duplicate of an engineering issue is the everyday version of this. Only remove the users who were actually moved, from both the source and the previous survivor, and add a cross-team test for each case.

packages/core/src/work/issue-service.ts
```diff
@@ -2663,3 +2664,584 @@ export async function findDuplicateIssues(
...
+      .where(
+        and(
+          eq(schema.issueRelation.organizationId, principal.organizationId),
+          eq(schema.issueRelation.issueId, source.id),
+          eq(schema.issueRelation.type, 'duplicate_of'),
+        ),
+      )
+      .limit(1);
+
+    if (existingDuplicateOf !== undefined && existingDuplicateOf.relatedIssueId === target.id) {
```
imshashank: [P2] This returns success with no actions whenever any duplicate_of row to the target exists. `setRelation` writes exactly that row without canceling the issue or moving anyone, and it is still offered in the Link issue picker, `POST /api/issues/:id/relations` and MCP `set_relation`. Every duplicate link already in production was made that way. Repro: link A "Duplicate of" B, then use Shift+M, the REST route or `mark_issue_duplicate` for the same pair. The result is 0 actions, A stays in Triage with `canceledAt` null, the subscribers stay on A, and the UI toasts "Marked issue as duplicate". A marked duplicate that was later reopened hits the same return.

Related: `.limit(1)` has no ordering, so with two duplicate_of rows (possible through `setRelation`) whether this fires is arbitrary. After unlinking a duplicate pair with the X button, a later mark finds no previous survivor, so the subscribers moved by the first mark stay on the old survivor. Only short-circuit when the issue is actually canceled and the single existing row points at the target. Now that there is a dedicated flow, consider refusing duplicate_of and duplicated_by in `setRelation`.

packages/mcp-server/src/tools/issues.ts
```diff
@@ -642,6 +643,37 @@ function registerRemoveRelation(server: McpServer, principal: Principal): void {
...
 
+function registerMarkIssueDuplicate(server: McpServer, principal: Principal): void {
+  defineTool(
+    server,
+    {
+      name: 'mark_issue_duplicate',
+      title: 'Mark an issue as duplicate',
+      description:
+        'Mark an issue as a duplicate of another issue. Moves the duplicate issue to the team canceled state, records the duplicate_of relation, and transfers subscribers to the survivor issue.',
+      readOnly: false,
```
imshashank, line 654: [P3] `mark_issue_duplicate` cancels the issue, deletes its subscriptions and replaces earlier duplicate links, so it should carry `destructive: true`, like `remove_relation` above it. Once the early return is fixed it is also safe to repeat, so `idempotent: true` fits too.

imshashank (changes requested):
Re-reviewed 0c92c0db. The two cases from my last review now have tests, and both tests fail against the previous head, so the independent-subscriber and explicit-unsubscribe fixes are real for the exact sequences they cover. Checking the rest of the subscription contract turned up more. Everything below was reproduced with temporary tests against this head merged with current main.

Blocking (details inline):
1. The provenance rows show up in the issue timeline and credit the wrong person.
2. A second duplicate merged into the same survivor loses its subscriber when the first duplicate is re-pointed.
3. Transfer rows never go stale, so a re-point still deletes subscriptions users chose after the transfer, and can put users back on issues they left.
4. Subscribers who cannot read the survivor are dropped from both issues.
5. Mark as duplicate silently does nothing when a duplicate_of link already exists, which covers every duplicate link in production today.

Items 1 to 3 all come from using `issue_activity` as the provenance store. Two designs I'd accept: keep provenance on `issue_subscription` itself (a nullable column naming the duplicate that brought the user there, cleared whenever the user subscribes or unsubscribes on their own), or drop the "take them off the old survivor" step so a re-point only ever adds subscribers and never removes any, which removes the need for provenance entirely. Your call, but the current approach needs another patch for every sequence we try.

Tests to add with the fix. Each of these can be broken today with every test still green:
- A guest (issue:read only) is refused by `markAsDuplicate`, the REST route and the MCP tool. Deleting `assertCan(principal, 'issue:update')` at issue-service.ts:3048 passes the whole suite.
- A survivor in a team some subscribers cannot read (item 4).
- The sync actions: subscription inserts go to the user scope, and the REST route publishes. Removing `await publish(result.actions)` from the route passes.
- A second identical call, and a history longer than one re-point (A to B to C and back to B).
- The preference for a canceled state named Duplicate, using a team that has both Canceled and Duplicate.

CI is green on this head. The branch is 4 commits behind main and merges cleanly.

packages/mcp-server/src/tools/issues.ts
Pallavikumarimdb: [3014e76](https://github.com/Noveum/orbit/pull/440/commits/3014e76c93dc676e5e344fd4023649b1d924a940)

packages/core/src/work/issue-service.ts
Pallavikumarimdb: [3014e76](https://github.com/Noveum/orbit/pull/440/commits/3014e76c93dc676e5e344fd4023649b1d924a940)

packages/core/src/work/issue-service.ts
Pallavikumarimdb: [3014e76](https://github.com/Noveum/orbit/pull/440/commits/3014e76c93dc676e5e344fd4023649b1d924a940)

packages/core/src/work/issue-service.ts
Pallavikumarimdb: [3014e76](https://github.com/Noveum/orbit/pull/440/commits/3014e76c93dc676e5e344fd4023649b1d924a940)

packages/core/src/work/issue-service.ts
Pallavikumarimdb: [3014e76](https://github.com/Noveum/orbit/pull/440/commits/3014e76c93dc676e5e344fd4023649b1d924a940)

packages/core/src/work/issue-service.ts
Pallavikumarimdb: [3014e76](https://github.com/Noveum/orbit/pull/440/commits/3014e76c93dc676e5e344fd4023649b1d924a940)

Pallavikumarimdb:
> Re-reviewed [0c92c0d](https://github.com/Noveum/orbit/commit/0c92c0db98dbde4d02622c53e82c933435de8d0c). The two cases from my last review now have tests, and both tests fail against the previous head, so the independent-subscriber and explicit-unsubscribe fixes are real for the exact sequences they cover. Checking the rest of the subscription contract turned up more. Everything below was reproduced with temporary tests against this head merged with current main.
> 
> Blocking (details inline):
> 
>     1. The provenance rows show up in the issue timeline and credit the wrong person.
> 
>     2. A second duplicate merged into the same survivor loses its subscriber when the first duplicate is re-pointed.
> 
>     3. Transfer rows never go stale, so a re-point still deletes subscriptions users chose after the transfer, and can put users back on issues they left.
> 
>     4. Subscribers who cannot read the survivor are dropped from both issues.
> 
>     5. Mark as duplicate silently does nothing when a duplicate_of link already exists, which covers every duplicate link in production today.
> 
> 
> Items 1 to 3 all come from using `issue_activity` as the provenance store. Two designs I'd accept: keep provenance on `issue_subscription` itself (a nullable column naming the duplicate that brought the user there, cleared whenever the user subscribes or unsubscribes on their own), or drop the "take them off the old survivor" step so a re-point only ever adds subscribers and never removes any, which removes the need for provenance entirely. Your call, but the current approach needs another patch for every sequence we try.
> 
> Tests to add with the fix. Each of these can be broken today with every test still green:
> 
>     * A guest (issue:read only) is refused by `markAsDuplicate`, the REST route and the MCP tool. Deleting `assertCan(principal, 'issue:update')` at issue-service.ts:3048 passes the whole suite.
> 
>     * A survivor in a team some subscribers cannot read (item 4).
> 
>     * The sync actions: subscription inserts go to the user scope, and the REST route publishes. Removing `await publish(result.actions)` from the route passes.
> 
>     * A second identical call, and a history longer than one re-point (A to B to C and back to B).
> 
>     * The preference for a canceled state named Duplicate, using a team that has both Canceled and Duplicate.
> 
> 
> CI is green on this head. The branch is 4 commits behind main and merges cleanly.

Simplified duplicate subscriber transfer to add-only and enforce permissions:

- Transfer duplicate subscribers to survivors as add-only without removing source or previous survivor subscriptions
- Remove subscription_transfer activity log entries to prevent timeline pollution and incorrect attribution
- Enforce issue:update permission check in markAsDuplicate for guest role refusal
- Disallow duplicate_of and duplicated_by relation types in setRelation
- Prefer team canceled state named 'Duplicate' when canceling duplicate issues
- Mark MCP tool mark_issue_duplicate as destructive and idempotent

imshashank (changes requested):
The test updates do not fix the existing entry points: Link issue still offers duplicate_of and duplicated_by through setRelation, which now rejects both. The MCP relate_issues tool still advertises the same unsupported choices. The new Mark duplicate picker also excludes every already-linked issue, so an issue marked related cannot be selected there even though the hotkey path allows it. Please route or remove the old duplicate choices, allow valid linked survivors, and add UI and MCP regressions.

Pallavikumarimdb:
> The test updates do not fix the existing entry points: Link issue still offers duplicate_of and duplicated_by through setRelation, which now rejects both. The MCP relate_issues tool still advertises the same unsupported choices. The new Mark duplicate picker also excludes every already-linked issue, so an issue marked related cannot be selected there even though the hotkey path allows it. Please route or remove the old duplicate choices, allow valid linked survivors, and add UI and MCP regressions.


Fixed in the latest commit!

- **`setRelation` Routing**: `setRelation` now routes `duplicate_of` and `duplicated_by` calls directly to `markAsDuplicate`. Generic entry points (like the MCP `set_relation` tool and UI Link dialog) now execute the full duplicate marking workflow cleanly.
- **Linked Survivor Selection**: Updated the UI `mark-duplicate-picker` to exclude only `issue.id`, allowing already-linked issues (e.g. `related`) to be selected as duplicate survivors.
- **Regressions Added**: Added core service, MCP server, and UI component tests covering duplicate routing through `setRelation` and selecting linked issues in the duplicate picker.

imshashank (approved):
The duplicate relation entry points and already-linked survivor picker are fixed. I checked the updated core paths and UI behavior; 30 core tests, 8 relation UI tests, and 3 separately run API tests pass. This replaces my earlier change request for those defects. Bring in current main and rerun all checks before merging.

imshashank (approved):
Rechecked the branch after merging current main. Fresh lint, type, migration, build, unit/integration, and both browser suites all pass on this commit, and no review threads remain open. The previous code findings are fixed, so this is ready to merge.

PR #442, opened 2026-09-08
Fix document permissions and redesign docs sharing and editing

pulkitxm:
Make documents private by default, unify sharing into three audiences, and provide a compact, resizable writing and reading experience.


imshashank:
Status check: this is moving fast, appreciate the quick turnaround on both CodeRabbit passes already landed, the workspace-visibility mapping and anonymous-access test fixes in 46b00b3, and the legacy `team` -> `workspace` normalization plus the `publishToken: null` assertion from the most recent pass.

One thing worth flagging before this goes further: `Unit and integration tests` failed on the latest commit (cdfd5ce9). The `@orbit/web` suite itself finished clean, 2662 pass, 0 fail, so whatever failed is in a different workspace package and it's past the tail of the log I could pull here. Worth checking the Actions run directly rather than assuming it's the same parallel-database-contention flake we've seen elsewhere, since I can't confirm that from what I have.

CodeRabbit is also sitting at "Review rate limited" again on this head, so per the repo's review policy I'd wait for that to complete before merging, both bots need to finish before this is ready either way.

---

imshashank (approved):
Read the policy change and traced it through HTTP, storage, realtime scopes, notification audience and the access-request service. They all agree now: `canReadDoc` no longer trusts the admin role, every team grant lookup derives from `team_member` rather than the principal's expanded team list, `listDocAccess` and visibility changes are author-only, and a revoked reader gets a content-free `revoked` action while `revalidateMutableResourceScopes` drops the held scope. No escalation path found. CI green, all 11 bot threads resolved, CodeRabbit completed on the head. Approving.

One design consequence to decide on, either in this PR or as an explicit follow-up issue:
- A private doc whose author leaves the workspace (or is deactivated) becomes unreadable, unshareable and undeletable by everyone, admins included, because `deleteDoc` goes through `loadReadableDoc` before it reaches the admin branch. Either an orphan reclaim flow or a narrow, audited admin escape hatch is needed eventually. Not blocking, but please say which way you want to go.

Small follow-ups, not blocking:
- `content.ts` keeps the `visibility` column default at `workspace` while `docCreateSchema` now defaults to `private`. Every service path passes the parsed value so behaviour is right, but the column default now lies.
- `realtime-server/src/auth.ts` still short-circuits the team branch on `principal.teamIds.length === 0`; redundant now that the subquery uses `team_member`.
- `ResizableRail` is imported by the app shell from a feature folder; it belongs under `components/ui`.
- The turbopack disk cache flag and the keyboard provider change are unrelated to doc permissions.
- Worth stating in the description: `members`, `link` and `public` docs now need an explicit write grant, where before any member with `doc:write` could edit them.

Updating the branch from main so checks run against the main it will land on, then merging when green.

---

Tanzeelsameen (approved):

PR #459, opened 2026-09-10
fix(db): reconcile schema objects against all migrations during baseline

### LINKED ISSUE #456: bug(db): migration-release reconciliation misses constraints from already-applied migrations
`packages/db/tests/migration-release.test.ts › database release > does not silently baseline a missing webhook ownership constraint` fails whenever exactly one new migration is pending and an earlier, already-applied migration is the one that actually owns the missing object.

Repro path found while investigating CI on #448: that PR adds `0027_friendly_bloodaxe.sql` (the `ai_usage` table). The test drops `webhook_delivery_processing_claim_check`, a constraint created by `0026_breezy_johnny_storm.sql`, then removes the latest migration from the ledger and runs `releaseDatabase()` expecting the constraint to be restored.

With only `0027` pending, `reconcileNotificationChecks()` (or the equivalent reconciliation step) is only given the pending migrations to work from. Since `0026` is already recorded as applied, the missing constraint it owns is invisible to reconciliation, and the constraint never gets recreated. This isn't specific to `0027` or to the AI provider work in #448, any PR that adds the next migration on top of `main` will trip the same test, since the failure depends on which migration is "latest" at merge time, not on anything the new migration itself does.

Root cause looks like reconciliation scoping itself to pending migrations only, when it needs to check the full set of declared schema objects regardless of which migration originally created them (consistent with how `bun run db:release` is documented to compare columns, types, defaults, indexes, foreign keys, etc. against the declared schema as a whole, not migration-by-migration).

Flagged from #448's thread: https://github.com/Noveum/orbit/pull/448#issuecomment-5610595793 (original diagnosis by @Pallavikumarimdb: https://github.com/Noveum/orbit/pull/448#issuecomment-5606023171).

Ayaan2907:
Title: fix(db): reconcile schema objects against all migrations during baseline

## What this changes
baselineLedger() reconciles the notification/webhook CHECK constraints and the
deferred audit triggers against the full declared migration set instead of only
the pending migrations. Ledger insertion and historical data backfills stay
scoped to pending.

## Why
Closes #456
During a partial baseline, a constraint or trigger owned by an already-applied
migration was invisible to reconciliation (it only saw pending migrations) and
never recreated. check-drift compares tables/columns/PKs/indexes/FKs/enums but
not CHECK constraints or triggers, so releaseDatabase() baselined silently
instead of restoring the object. The failure depends on which migration is
latest at merge time, so any PR adding the next migration on top of main trips
it (surfaced on #448).

## How you know it works
New test "reconciles objects owned by already-applied migrations when a later
migration is pending" synthesizes a trailing migration so the owners are all
applied while a later migration is pending. It fails before the change
(constraint not restored) and passes after. migration-release suite 14/14,
@orbit/db 291/291, typecheck, comment policy and Biome all clean.

## Checklist
- [x] No any, no non-null assertions
- [x] db:check-drift behavior unchanged (drift model untouched)


imshashank (commented):
Reviewed head 4d1490f9. The change correctly reconciles schema artifacts against all declared migrations while leaving ledger insertion and historical data backfills scoped to pending migrations. The synthetic trailing-migration test exercises the reported partial-baseline case; all 14 migration-release tests pass locally. No additional blocking code finding.

Updated maintainer direction: manual review replaces Greptile and CodeRabbit availability requirements. The earlier bot-quota merge hold is withdrawn. Merge readiness will be decided from reviewed code, addressed findings, and successful checks against current main. No separate requester screen-review hold is being imposed under this direction.

imshashank (approved):
Manual review approved. The migration-baseline reconciliation fix and its regression test remain correct after bringing in current main. The 14 focused migration-release tests passed during review. No unresolved code findings. Bot availability is not a merge prerequisite under the maintainer direction; merge after current-head CI passes.

PR #476, opened 2026-09-16
fix(docs): give an html page an artifact link inside the workspace

pulkitxm:
Restore the workspace-only artifact link the sharing redesign dropped, keyed by doc id and gated by the doc's own read check.


imshashank (approved):
Traced the token lifecycle and the read gate end to end. A workspace or members token only resolves for a signed-in principal of the same organization, `private` and `team` never carry a token, the token rotates when a doc is pulled back from `link` or `public` so a leaked external URL dies, the `/h/` route redirects anonymous viewers to sign-in and serves under the sandbox CSP, `getPublishedDoc` and the sitemap still ignore workspace docs, and `docAnnouncement` keeps redacting the token on the socket. The backfill migration is mirrored in the release baseline path like the earlier data migrations. Tests cover every transition plus the e2e stranger redirect. CI green on the merged state, both CodeRabbit threads fixed and resolved. Approving.

Three optional nits:
- The share-dialog hint "The page on its own, read only" reads html-specific but shows for markdown docs too.
- The migration mints 64-hex tokens via `gen_random_uuid()` while the app uses `newToken()`, so two token shapes coexist. Harmless, equality-matched only.
- The backfill SQL now lives in both the migration file and `migration-release.ts`. Consistent with the existing pattern, just noting it.

Worth stating in the description: the PR no longer adds an artifact route; it gives every workspace-shared doc a token instead, and a token circulated inside the workspace becomes the public URL the moment the owner publishes.

Greptile has not reviewed anything on this repo since its trial ended, so this merges on CodeRabbit plus CI.

---

PR #478, opened 2026-09-17
feat(settings): channel toggles for notifications, and clickable links in descriptions

Tanzeelsameen:
## What this changes

Notification settings lead with one toggle per channel instead of a 24 by 4 grid of checkboxes, with the per type picks moved behind a Customize panel on the channel they belong to. Separately, a link in an editable rich text surface is now clickable.

## Why

Two things reported off the same screen.

**The settings grid asked ninety six questions.** Twenty four notification types across four channels, every cell a checkbox, no default and no summary. The question a person actually has is which channels reach them, and only after that which of the twenty four types each one carries. The grid answered neither at a glance.

**Links in a description were dead.** A description is an editable ProseMirror surface, so clicking an anchor placed the caret instead of following the link. The link extension was also configured with `openOnClick: false`, so nothing opened it. Read only surfaces were never affected, which is why this only showed up on issues and docs you can edit.

## Scope of the link change

The report was about issue descriptions, but the fix is not scoped to them, and that is deliberate. `LinkClick` goes into `editorExtensions()`, which feeds every editable rich text surface: the doc editor, the comment composer, quick create and the project update composer. A link in any of them now opens on a plain left click, including the first click of a double click to select a word, while the caret still lands. Shift click and a drag selection are unaffected, because ProseMirror does not call `handleClick` for those.

Scoping it to descriptions would have meant a second extension list and a surface where a link stays dead for no reason a user could explain. This is the same behaviour as Tiptap's own `openOnClick` default, with a stricter protocol list and an explicit `noopener`.

An in page link such as `[Setup](#setup)` is left to the page rather than opened in a second tab: `openableHref` returns null when the href carries a fragment delimiter and resolves to the current document, which covers a bare `#` as well as a named target.

## How you know it works

Automated, all failing without the change:

- `notification-channels.test.tsx`, 27 tests. Beyond the shape of the screen (one master per channel, the panel revealing exactly the 24 type switches, live counts, Slack kept out entirely while the capability is off), each behavioural test was written against a deliberate break in the source and confirmed red first: a group All that behaves like None, a type switch that can only ever go off, a master that cannot turn on a channel that loaded fully off, a channel edit landing while a save is in flight, a save that fails, a Slack DM lock narrowed to one of its three states, both routes by which a channel snapshot can outlive the state it was taken from, and the snapshot restore under a state updater that runs twice.
- `notification-groups.test.ts`: every notification type appears in exactly one group and carries a label. This is what stops a new entry in `NOTIFICATION_TYPES` silently disappearing from the UI.
- `link-click.test.tsx`, 13 tests. The guards in the click handler are covered the same way: a read only surface, a click on plain text while the document holds a link elsewhere, a button other than the primary one, and a relative href proving the resolved URL rather than the raw attribute reaches `window.open`. Plus the unit cases for `javascript:`, `data:`, `file:` and `vbscript:`, and for a same document fragment, empty or named.

These tests do not touch the network. happy-dom activates an anchor natively on click and, for `target="_blank"`, navigates through its own internal window, which is not the mocked `window.open`, so an earlier version of this file sent a real request to github.com on every run. A capture phase listener now blocks navigation for the duration of the file. ProseMirror runs `handleClick` on mouseup, so the plugin path still executes.

Manual, against `bun run dev` with the seeded workspace, driven through Chromium: 16 assertions covering both changes, including preferences surviving a save and reload, the Slack DM card locking itself with its reason, and the opened tab reporting `window.opener === null`.

## Screenshots

Four toggles where there were ninety six checkboxes. Slack DM is locked here because this workspace has no Slack connection, and it says so instead of showing a count.

| Dark | Light |
| --- | --- |
|  |  |

Email customised. Six groups with per group counts, `Reaction` and the whole Workspace group switched off, and the channel summary tracking it live at "On for 21 of 24 notifications".

| Dark | Light |
| --- | --- |
|  |  |

The before is the 24 by 4 checkbox grid this replaces, still visible in the diff of `notification-matrix.tsx`.

## Checklist

- [x] `bun run verify` is green, every check

The two database items do not apply. No schema change, so there is nothing to release or drift check, and no documented behaviour, configuration or setup changed.

Every CI job is green on the current head, `cc2cd6b`, including `Unit and integration tests`, `Lint, comments, types`, `Build`, both end to end jobs and both CodeQL analyses. Nothing was skipped or quarantined to get there.

For the record, since it was noted on an earlier revision of this description: the full `bun run verify` in the sandbox this was developed in was red on both this branch and `main`, with `SyntaxError: Export named 'redirect' not found in module 'next/navigation.js'` once 300 plus test files share a process. CI has now settled that as an artefact of that environment rather than anything in this diff.

**Nothing migrates.** The stored shape is untouched: the same channel by type rows go to `PUT /api/notifications/preferences`, and `disabledKeys` still seeds the UI. A channel master is derived from its rows, it is not a new stored flag. That is deliberate, because it means an existing account's preferences read back exactly as before.

**Inbox stayed a real channel, against the mockup.** The design had Inbox locked as "Always on", matching the old page copy. It is not true: `planFor` in `packages/services/src/notifications/index.ts` reads the inbox preference like any other and suppresses the inbox row when it is off. Locking it in the UI would have stranded existing `inbox:*` rows behind a control claiming they did not exist. Inbox now gets the same toggle as the rest, and the page copy no longer promises something the server does not do.

**The link fix does not use the extension's own `openOnClick`.** Setting it to `true` would be one word, but it calls `window.open(href, target)`, and `window.open` does not inherit the anchor's `rel`, so the opened tab keeps a live `window.opener` back into Orbit. Descriptions are user authored and shared across a workspace, so the small plugin in `link-click.ts` passes `noopener,noreferrer` explicitly and allows only `http`, `https` and `mailto`. A read only surface returns early and is left to the browser.

**One shared component moved.** `Switch` had its unchecked thumb on `bg-surface`, which is very close to the track colour in dark mode. It was tolerable when a switch was a secondary control on this page, but the switch is now the primary one, so the unchecked thumb moved to `bg-muted`. This affects every switch in the app, which is the intent, and the off rows in the second pair of screenshots are what it buys.

**Dropped from the mockup:** a "Turn all on" and "Turn all off" pair in the panel toolbar. The channel master sits directly above it and does the same job, so they were redundant. Per group All and None stayed, since those are not reachable any other way.

**Review findings fixed.** From the first round, both inherited verbatim from `notification-matrix.tsx` rather than introduced here: the quiet hours copy promised to hold Slack DMs in the `unavailable`, `unmapped` and `reauthorize` states where that channel delivers nothing, and the quiet hours and urgent bypass controls bypassed the reset the channel toggles go through, leaving the saved confirmation on screen over an unpersisted change. Then one introduced here: the controls stay interactive while the save request is in flight and the payload is the snapshot from when it started, so an edit landing mid request was unpersisted while the note still read "Notification preferences saved". An edit revision now guards that note.

From the second round, one real bug and two accessibility fixes. The remembered snapshot a channel master writes when it goes off was never cleared, so it outlived the off and on pair it belonged to and replayed stale picks over choices made in between; it is now consumed when it is restored, with the reported repro kept as a test. Every notification row is now a `label` bound to its switch, so the row is the hit target rather than the 18 pixel switch alone, and the All and None buttons start their accessible names with their visible text, which WCAG 2.5.3 requires and speech input needs.

From the third round, the same document guard turned out to be one case short: a bare `#` resolves with an empty hash, so it still opened a second tab of the current page. It now looks for the delimiter in the href the anchor carries. The round trip test also gained an assertion on `quietHoursEnd`, which nothing was checking, and the three suites this branch adds import through `@/`.

From the fourth round, a second route into the snapshot bug that consuming it on restore did not close. The master writing the snapshot is not the only thing that can change a channel: the type switches and the group buttons edit the same keys without going through `setChannelEnabled`, so a channel taken back to fully off through those controls kept a snapshot from before the master ever went off, and switching the master on restored it over what the user had chosen since. Any edit other than the master turning the channel off now discards the snapshot, so it can only belong to the off and on pair that created it.

From the fifth round, the snapshot fix itself was not safe to run twice. Reading and clearing the snapshot happened inside the function handed to `setDisabled`, which React invokes twice in development and takes the second result from, so the second pass found the snapshot gone and turned every type on. With `reactStrictMode` on, the plainest sequence hit it: channel off, Save, channel on came back as everything enabled, which is the headline behaviour of this PR broken for anyone running the app locally. The read and the clear now happen in the handler, matching `toggleType` and `setGroupEnabled`, and the updater only edits the set it is given. All three routes into the snapshot now have their own regression test, and each fails only against its own fix.

**One finding declined,** with reasoning on its thread: replacing the `Collapsible` around the Customize panel with conditional rendering, because it transitions `grid-template-rows`. The component is pre-existing and untouched by this diff, and four other surfaces use it, so the fix belongs in the component rather than at one call site. It is being tracked separately.

**One nitpick declined and withdrawn:** using `@/` for `./link-click.ts` inside `extensions.ts`. That file already imports two siblings relatively, `src` holds 495 sibling relative imports, and `src/features/docs/editor/` has no same directory alias imports at all, so the repository reads the `@/` rule as being about reaching across the tree. The same nitpick was correct on the test files, which reached four levels up into `src`, and was taken there.

**Tracked separately, found during review and not changed here:** quiet hours are evaluated in `user.timezone`, which defaults to `UTC`, is `input: false` at sign up and is only ever set by typing a zone into Profile. So the card's "in your local time" is not true for anyone who has never opened that page, and the default 18:00 to 09:00 window is 23:30 to 14:30 in India. The fix is to detect the zone after the account exists, not to reword the card.

**The screenshots are committed** to `docs/assets/screenshots/` under the gallery's existing `notification-*` naming, captured at its documented 1680 by 1000 viewport at a device scale factor of two, per the note in that folder's README that feature specific captures may come from the feature's own browser verification. Drop `a84690a` if you would rather they were not in the tree; the images above are pinned to that commit, so they would stop rendering here.

**The branch was rewritten once,** to correct the author and committer on every commit. No content changed in that rewrite.


apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,404 @@
...
+          request stay in one Slack thread.
+        </p>
+      )}
+
+      <div className="flex flex-col gap-3 rounded-lg border border-border bg-surface p-4">
+        <label htmlFor="quiet-hours" className="flex items-center justify-between gap-3">
+          <span className="flex flex-col">
+            <span className="font-medium text-dense text-text">Quiet hours</span>
+            <span className="text-muted text-xs">
+              {props.slackDm === 'disabled'
```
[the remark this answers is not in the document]
Tanzeelsameen: Correct, fixed in 1d95c30. Checked it against the delivery side before changing the copy: `planFor` computes `deferred` from `emailEnabled || slackDmEnabled`, and `slackDmEnabled` is false for `unavailable`, `unmapped` and `reauthorize` alike, so those three hold nothing over Slack. The conditional is now keyed on `=== 'available'` as you proposed.

Covered by two tests in `notification-channels.test.tsx`, one asserting the email-only wording under `unavailable` and one asserting the Slack wording survives under `available`, so the branch cannot silently flip back.

Worth noting this predates the PR: the same conditional was in `notification-matrix.tsx`. It is in this diff now, so it is fixed here.

---

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,404 @@
...
+                ? 'Email is held until the window ends, in your local time.'
+                : 'Email and Slack DMs are held until the window ends, in your local time.'}
+            </span>
+          </span>
+          <Switch
+            id="quiet-hours"
+            checked={quietHoursEnabled}
+            onCheckedChange={setQuietHoursEnabled}
```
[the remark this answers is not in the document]
Tanzeelsameen: Correct, fixed in 1d95c30. The channel and type toggles already cleared the note through the `apply()` helper; these four handlers set state directly and bypassed it, so saving and then editing the quiet-hours window left the confirmation on screen over an unpersisted value.

Rather than inline the reset four times I routed them through a small `edit()` wrapper that clears the note and then delegates, which keeps the JSX readable and makes it harder to add a fifth control that forgets:

```tsx
function edit<T>(set: (value: T) => void): (value: T) => void {
  return (value) => {
    setSaved(false);
    set(value);
  };
}
```

Two tests cover it, one for quiet hours and one for urgent bypass, each saving first and then asserting the note is gone after a change. I confirmed they fail against the previous code rather than assuming.

Same note as the other thread: this behaviour came over from `notification-matrix.tsx` and was not introduced here, but it is in this diff so it is fixed here.

---

Tanzeelsameen:
@coderabbitai review

The previous run hit the included-review limit before reaching `1d95c30`, so the two fixes it asked for are still unreviewed and the commit status reads "Review rate limited" rather than a real result.

Tanzeelsameen:
@coderabbitai full review

The earlier request returned a note that reviews were available again, but no review ran against `1d95c30`, and the commit status still carries the "Review rate limited" result from the previous head.

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,411 @@
...
+        method: 'PUT',
+        body: {
+          preferences,
+          quietHoursEnabled,
+          quietHoursStart,
+          quietHoursEnd,
+          urgentBypassEnabled,
+        },
+      });
+      setSaved(true);
```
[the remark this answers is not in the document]
Tanzeelsameen: Correct, and reachable. Fixed in 8a578a8.

Confirmed against the code before changing it: only the Save button carries `disabled={pending}`, so every channel, type and quiet hours control stays live for the whole request, and `preferences` is built from the `disabled` snapshot captured when `save()` ran. A toggle landing in that window persists nothing and still ends on "Notification preferences saved."

Implemented as proposed. `editRevision` advances in both `apply` and `edit`, `save` captures it before the request, and the note is only shown when the revision that started the request is still current:

```tsx
if (editRevision.current === revision) setSaved(true);
```

A ref rather than state on purpose: this must not re-render, and it has to be readable inside the closure after the await without going stale.

Covered by a test that holds the request open, toggles quiet hours while it is in flight, then releases it and asserts the note never appears. Reverting the one-line guard turns it red, so it fails for the reason it claims to.

Worth noting the payload is still the older snapshot, which is correct: the user's next save sends the newer state, and the note no longer misreports the older one as current.

---

apps/web/tests/features/docs/editor/link-click.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+function renderDescription(value: string) {
+  render(<RichTextEditor value={value} onChange={mock()} ariaLabel="Issue description" />);
+}
+
+describe('opening links from an editable description', () => {
+  it('opens a link in a new tab with no reference back to Orbit', async () => {
+    const user = userEvent.setup();
+    renderDescription('PR: [936](https://github.com/Noveum/orbit/pull/936)');
+
+    await user.click(await screen.findByRole('link', { name: '936' }));
```
imshashank, line 41: This click sends a real HTTPS request to github.com on every run. happy-dom activates the anchor natively on click, and for `target="_blank"` it opens a page and navigates to the href through its own internal window, which is not the `window.open` you mocked. I ran this file with a preload that logs `node:https` requests and got `outbound https.request -> https://github.com/Noveum/orbit/pull/936`, with all 6 tests passing.

Unit tests must not touch the network: it makes CI depend on github.com and leaves an unawaited navigation running after the test ends.

Register a capture phase `click` listener on `document` in `beforeEach` that calls `preventDefault()`, and remove it in `afterEach`. ProseMirror runs `handleClick` on mouseup, before the click event, so the plugin path still executes. I tried exactly that: 0 outbound requests and the `toHaveBeenCalledWith` assertion still passes.

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,415 @@
...
+  function setChannelEnabled(channel: NotificationChannel, enabled: boolean): void {
+    if (isLocked(channel)) return;
+    const keys = channelKeys(channel);
+    apply((next) => {
+      if (!enabled) {
+        remembered.current.set(channel, new Set(keys.filter((key) => next.has(key))));
+        for (const key of keys) next.add(key);
+        return;
+      }
+      const previous = remembered.current.get(channel);
```
imshashank, line 157: The snapshot is written only when the master goes off and is never cleared, so it outlives the off and on pair it belongs to and later overwrites an explicit choice.

Repro: load with `email:mention` disabled. Email master off, then on (restored, correct). Open Customize and turn Email for Mention on, so all 24 are on. Click None on all six groups, so the master reads off. Turn the master on. Mention comes back disabled and the PUT carries `email:mention enabled=false`, although the user had just enabled it. It repeats on every later round until the next master off rewrites the snapshot.

The master off, on, edit, off, on sequence is fine, because off rewrites the snapshot. It only breaks when the channel reaches fully off through group None or the per type switches.

Fix: consume the snapshot when you restore from it.

```ts
const previous = remembered.current.get(channel);
remembered.current.delete(channel);
```

Please add the repro above as a test. It currently saves `["email:mention"]` and should save `[]`.

Separately, `previous.size < keys.length` on the next line can never be false: the master only fires off while at least one type is on, so a snapshot holds at most 23 keys. Drop the guard or say what it protects.

apps/web/tests/features/settings/notification-channels.test.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+    await waitFor(() => {
+      expect(sentBody).not.toBeNull();
+    });
+    expect(disabledAfterSave()).toEqual(
+      NOTIFICATION_TYPES.map((type) => channelTypeKey('push', type)).sort(),
+    );
+  });
+
+  it('turns one group on or off without touching the rest of the channel', async () => {
```
imshashank, line 136: This test is titled "turns one group on or off" but only ever clicks None. I made each of these changes to `notification-channels.tsx` one at a time and the whole file stayed green:

- `setGroupEnabled` always adds the key, so All behaves as None.
- `toggleType` always adds the key, so a type that loaded off can never be turned on.
- `setChannelEnabled` returns early when there is no snapshot, so a channel that loaded fully off can never be turned on by its master. The only re-enable path covered is master off then on with exactly one earlier pick.
- `apply()` no longer bumps `editRevision`. The in flight test only edits quiet hours, which goes through `edit()`, so a channel toggle landing mid request is not covered. That is the primary control on this screen.
- `apply()` no longer calls `setSaved(false)`. Only quiet hours and urgent bypass are checked for withdrawing the saved note.
- `catch` swallows the error, or `setPending(false)` moves out of `finally`. There is no failed save test at all, so the button can stick on Saving after a 500 with nothing red.
- `isLocked` and the skip in `save()` narrowed to `=== 'unavailable'`. Only that one locked state is rendered, so `unmapped` and `reauthorize` could start sending `slack_dm` rows.

Please add: group All, a per type switch turned back on, master on for a channel that loaded fully off, a channel toggle during an in flight save, a failed save (error shown, button usable again), and the locked assertions parametrised over `unmapped` and `reauthorize` as well.

Also, the old round trip test seeded a stale `slack:reaction` key to prove it never reaches the payload. The new one seeds `[]`. Please keep that seed.

apps/web/src/features/docs/editor/link-click.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const LinkClick = Extension.create({
+  name: 'orbitLinkClick',
+  addProseMirrorPlugins() {
+    return [
+      new Plugin({
+        key: new PluginKey('orbitLinkClick'),
+        props: {
+          handleClick: (view, _pos, event) => {
+            if (!view.editable || event.button !== 0) return false;
```
imshashank, line 37: None of the guards in this handler are tested. Each of these leaves `link-click.test.tsx` green:

- removing `!view.editable`, which is the whole "a read only surface is left to the browser" claim
- removing `event.button !== 0`, so a middle or right click opens it too
- replacing `target.closest('a')` with `view.dom.querySelector('a')`, so a click anywhere in a description opens its first link. The negative test renders a document with no link in it, so it cannot notice.
- passing `anchor.getAttribute('href')` straight to `window.open`. `openableHref` is unit tested, but nothing proves the handler uses it.

Three more cases cover all of it: an `editable={false}` render where the click records no open, a click on plain text that sits next to a link, and a relative href asserting the absolute URL reaches `window.open`.

apps/web/src/features/docs/editor/link-click.ts
```diff
@@ -0,0 +1,38 @@
...
+            const target = event.target;
+            if (!(target instanceof Element)) return false;
+            const anchor = target.closest('a');
+            if (anchor === null || !view.dom.contains(anchor)) return false;
+            const href = openableHref(anchor.getAttribute('href'), view.dom.baseURI);
```
imshashank, line 42: Minor. An in page link such as `[Setup](#setup)` now opens a second tab of the same document. It resolves against `baseURI` to the current URL plus a fragment and passes the protocol check. On `main` the click did nothing. The read view does resolve hash links (heading ids plus `use-hash-scroll`), so an author can write one.

Return null from `openableHref` when the resolved URL differs from `base` only by its fragment, and add the case to the unit tests.

apps/web/src/features/docs/editor/extensions.ts
```diff
@@ -110,6 +111,7 @@ export function editorExtensions(handler: MenuKeyHandlerRef, placeholder = '') {
     Callout,
     ToggleSummary,
     ToggleBlock,
+    LinkClick,
```
imshashank, line 114: Scope note rather than a defect. `editorExtensions()` feeds every editable surface, so this changes the doc editor, the comment composer, quick create and the project update composer, not only descriptions as the title says. In all of them a plain click inside link text, including the first click of a double click to select a word, now opens a tab while the caret lands. Shift click and a drag selection do not, because ProseMirror never calls `handleClick` for those.

That is the same behaviour as Tiptap's own `openOnClick` default, with a stricter protocol list and an explicit `noopener`, so I am fine with it. Please state the real scope in the description so it reads as a decision and not a side effect.

apps/web/src/features/settings/settings-skeletons.tsx
```diff
@@ -267,42 +267,27 @@ export function MembersSettingsSkeleton() {
   );
 }
 
-const MATRIX_ROWS = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h'];
-const MATRIX_CHANNELS = ['inbox', 'email', 'push'];
+const SKELETON_CHANNELS = ['inbox', 'email', 'slack_dm', 'push'];
```
imshashank: Minor. The skeleton always draws four cards including `slack_dm`, but `SLACK_ENABLED` defaults to false and the page then renders three, so quiet hours and the Save button jump up by one card when the page resolves. The old `MATRIX_CHANNELS` had three for that reason. Use three here.

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,415 @@
...
+                          return (
+                            <div
+                              key={type}
+                              className={cn(
+                                'flex items-center justify-between gap-3 rounded-sm px-1.5 py-1.5 text-xs',
+                                on ? 'text-secondary' : 'text-muted',
+                                rowHover,
+                              )}
+                            >
+                              {NOTIFICATION_TYPE_LABELS[type]}
```
imshashank, line 310: Minor. The row highlights on hover through `rowHover`, but the text is a bare text node, so clicking "Mention" does nothing and the only hit target for each of the 24 rows is the 18px switch. The quiet hours and urgent bypass rows below wrap theirs in `<label htmlFor>`. Do the same here (the Radix switch accepts an `id`), or drop `rowHover` so the row stops looking clickable.

Same family: the All and None buttons have accessible names that do not contain their visible text ("Turn on every Issues notification for Email" against "All"). That fails WCAG 2.5.3 Label in Name, and speech input users saying "click All" get no match. Start the label with the visible word, for example `All Issues notifications for Email`.

imshashank (changes requested):
I checked this out and ran it rather than reading the diff: the three new test files (25 pass), `tsc`, biome, the comment and bun-import checks, then the branch itself in Chromium against a seeded workspace.

What holds up:

- The save payload is identical to what the matrix sent on `main` for all five `slackDm` states, so nothing migrates, as you said. I saved from the browser and read the rows back: 72 rows, the three email picks and all 24 push rows disabled, and a reload shows the same state.
- Keeping Inbox a real channel is right. `planFor` does honour inbox rows, so the old "always keeps a copy" copy was already false on `main`.
- The `Switch` thumb change is an improvement everywhere, not only here. Unchecked thumb against track goes from 1.19 to 6.73 in light and from 1.07 to 5.47 in dark.
- `openableHref` is sound. 41 hostile hrefs (control characters inside the scheme, mixed case, `blob:`, `filesystem:`, protocol relative, userinfo tricks) all resolve to either null or a plain https link, and the opened tab has no `opener` and sends no referrer in Chromium.
- The three CodeRabbit threads are fixed in the code, not only resolved.

What needs to change before this merges:

1. The link click test makes a real HTTPS request to github.com on every run. Inline.
2. The remembered picks go stale and overwrite an explicit choice. One line fix, inline.
3. The tests miss the core re-enable paths. The description says the tests fail without the change, and that is true for the headline behaviours, but I broke the implementation in several ordinary ways and the suite stayed green. Inline, with the exact mutations.
4. Every commit on the branch is authored and committed under a tool identity instead of yours, and your replies on this PR carry a generated-by footer. That is hard rule 3 of the repo rules, no tool attribution anywhere. We squash, and GitHub turns a commit author who is not the PR author into a co-author trailer on `main`. Please reset the author on the six commits (`git rebase origin/main --exec 'git commit --amend --no-edit --reset-author'`, then force push) and drop the footer from replies.

Smaller things are inline and marked as such: the hash link case, the skeleton card count, and two accessibility nits.

Not for this PR, I will track them separately: turning a channel off is stored as one row per type that exists today, so a type added later arrives enabled on a channel the user switched off (this already happened three times on `main`); nothing in the delivery code reads the `push` channel, so that card configures something that never sends; and `Collapsible` animates `grid-template-rows`, which the motion rule forbids, but that component predates this branch.

apps/web/tests/features/docs/editor/link-click.test.tsx
Tanzeelsameen: Confirmed and fixed in `281acf9`. I reproduced it the same way before changing anything: a preload that patches `node:https` logged `outbound https.request -> https://github.com/Noveum/orbit/pull/936` with all tests passing, so the file was green while depending on github.com being reachable.

Registered exactly what you described, a capture phase `click` listener on `document` that calls `preventDefault()`, added in `beforeEach` and removed in `afterEach`. Zero outbound requests now, and the `toHaveBeenCalledWith` assertion still passes, so the plugin path is still the thing being exercised rather than the browser's.

Worth naming why it still works: ProseMirror derives `handleClick` from its own mouseup handling, so the plugin has already run and called `window.open` by the time the `click` event we cancel is dispatched. Cancelling the default action stops only happy-dom's native anchor activation.


---

apps/web/src/features/settings/notification-channels.tsx
Tanzeelsameen: Correct on both counts, and the repro is exact. Fixed in `2392ffe`.

I ran your sequence before changing the code and got what you said: `["email:mention"]` in the PUT after the user had just turned Mention on. Your diagnosis of why is right too, and it is the part I had missed: the master only ever writes the snapshot, so once a channel reaches fully off through group None or the per type switches, the snapshot on file belongs to an earlier off and on pair that has already been restored.

Consumed on restore, as proposed:

```ts
const previous = remembered.current.get(channel);
remembered.current.delete(channel);
```

The `previous.size < keys.length` guard is gone rather than documented, because you are right that it cannot be false. The master only fires off while at least one type is on, so a snapshot holds at most 23 of the 24 keys, and a guard that never fires is worse than no guard.

Your repro is now a test, `forgets an earlier channel snapshot once it has been restored`. It walks the whole sequence, master off, master on, Mention on, None on all six groups, master on, and asserts the payload disables nothing. Reverting the one line `delete` turns it red with exactly the diff you quoted, `- []` against `+ ["email:mention"]`.


---

apps/web/tests/features/settings/notification-channels.test.tsx
Tanzeelsameen: All seven added in `281acf9`, and I checked each one the way you did rather than assuming it bites. Every mutation below was applied to the source on its own, the suite run, and the source restored:

| Break | Caught by |
| --- | --- |
| `setGroupEnabled` always adds the key | `turns a group back on without touching the rest of the channel` |
| `toggleType` always adds the key | `turns one notification back on inside a channel` |
| `setChannelEnabled` returns early without a snapshot | `turns on a channel that was loaded with everything off` |
| `apply()` stops bumping `editRevision` | `does not call a stale save current when a channel changes while the request is in flight` |
| `apply()` stops calling `setSaved(false)` | `stops claiming preferences are saved once a channel changes again` |
| `catch` swallows the error | `reports a failed save and lets the save be tried again` |
| `setPending(false)` moves out of `finally` | same test, on the button still reading `Save preferences` |
| `isLocked` narrowed to `=== 'unavailable'` | the locked case, now `it.each` over `unmapped`, `reauthorize` and `unavailable` |

Two notes on the detail. The group test seeds the whole email channel off and then clicks All on one group, so it fails both if All behaves like None and if All leaks outside its group. The in flight test is a second one rather than a change to the existing one, because the quiet hours path through `edit()` and the channel path through `apply()` are separate increments and I wanted both pinned.

The `slack:reaction` seed is back in the round trip test. It is the right seed: it proves a stale key for a channel that is not rendered never reaches the payload, which is the case `visibleNotificationChannels` exists to handle.

The two older Slack DM tests are gone, folded into the parametrised one so the same assertions are not written twice.


---

apps/web/src/features/docs/editor/link-click.ts
Tanzeelsameen: Right on all four, including that two of them needed the test shaped a particular way to bite at all. Added in `281acf9`, each confirmed against the corresponding mutation:

- `does not take over a link while the editor is read only`, an `editable={false}` render where the click records no open. Removing `!view.editable` turns it red.
- `leaves text alone even when the document holds a link elsewhere`. Your point about the old negative test was the sharp one: it rendered a document with no link at all, so `view.dom.querySelector('a')` returned null and the mutation survived. It now renders two paragraphs, plain text and a link, and clicks the plain one.
- `opens the href the anchor resolves to, not the one it carries`, a relative `/issues/AM-1` asserting `new URL(href, document.baseURI).href` reaches `window.open`. Passing the raw attribute through survived every earlier test because the hrefs in them were already absolute.
- `leaves a click that is not the primary button to the browser`, a right click and a middle click through `user.pointer`. Removing `event.button !== 0` turns it red.

All twelve tests in the file now run with navigation cancelled, per the other thread, so none of this touches the network.


---

apps/web/src/features/docs/editor/link-click.ts
Tanzeelsameen: Agreed, and it is a real regression rather than a cosmetic one: on `main` the click did nothing, and this turned it into a duplicate tab. Fixed in `2392ffe`.

`openableHref` now returns null when the resolved URL matches the document on origin, pathname and search and differs only by fragment:

```ts
if (url.hash.length > 0 && sameDocument(url, base)) return null;
```

So `#setup` and the fully spelled out `https://orbit.test/issues/AM-215#setup` both fall through to the page, while `/issues/AM-1#summary`, `?tab=activity#summary` and `https://example.com/a#summary` still open, because those are other documents and the fragment is not what distinguishes them. Two unit tests cover each side of that line, and removing the guard turns them red.

One thing I did not do, in case it is what you would have wanted: it falls through rather than scrolling. Inside a contenteditable the browser will not follow the anchor either, so on an editable surface an in page link now does nothing, which is the `main` behaviour. Making it scroll means reaching for `use-hash-scroll` from an editor extension, which felt like more than this fix should carry. Happy to add it if you would rather the editable surface matched the read view.


---

apps/web/src/features/docs/editor/extensions.ts
Tanzeelsameen: Fair, the description undersold the reach. Added a `Scope of the link change` section that names all four surfaces, the double click first click behaviour and the shift click and drag exemptions, and states the choice rather than leaving it to be inferred:

> Scoping it to descriptions would have meant a second extension list and a surface where a link stays dead for no reason a user could explain.

It also records the comparison you drew, that this is Tiptap's own `openOnClick` default with a stricter protocol list and an explicit `noopener`, and the one place the behaviour now differs from that default: a same document fragment is left to the page rather than opened in a second tab, per your other thread.


---

apps/web/src/features/settings/settings-skeletons.tsx
Tanzeelsameen: Fixed in `2392ffe`, three cards. `SKELETON_CHANNELS` is now `['inbox', 'email', 'push']`.

You are right about why it has to be three rather than four: `slack_dm` is the only channel whose presence is conditional, and `SLACK_ENABLED` defaults to false, so four cards guarantees the jump on the default configuration while three is correct there and one card short only on a workspace that has Slack turned on. That is the trade the old `MATRIX_CHANNELS` was making too.


---

apps/web/src/features/settings/notification-channels.tsx
Tanzeelsameen: Both fixed in `2392ffe`, taking the `label` option rather than dropping `rowHover`. The row reads as clickable because it should be clickable, and an 18 pixel target repeated 24 times is the thing worth fixing.

Each row is now a `<label htmlFor>` around its text and switch, matching quiet hours and urgent bypass below. The id comes from `useId()` plus the channel and type, so two instances of the component on one page cannot collide:

```tsx
const switchId = `${panelId}-${channel}-${type}`;
```

The switch keeps its `aria-label`, which still wins the accessible name computation, so `Email for Mention` is unchanged for anyone querying by it.

On Label in Name, the two names now start with their visible word: `All Issues notifications for Email` and `None of the Issues notifications for Email`. "None of the" rather than a bare "None" so it reads as a sentence while still opening with the visible token.

A test asserts both, and that each row's `for` matches the id of the switch it wraps, so the binding cannot quietly break.


---

Tanzeelsameen:
@imshashank all nine items are addressed. Every thread has a reply and is resolved.

**Commit authorship.** You were right, and I had checked the wrong thing: I grepped the commit *messages* for attribution and never looked at the author and committer fields, so all six commits went up as `Claude <noreply@anthropic.com>`. Fixed exactly as you said:

```
git rebase origin/main --exec 'git commit --amend --no-edit --reset-author'
```

All eight commits are now `Tanzeelsameen <tanzeela@noveum.ai>` on both fields. The rewrite changed no content, and the screenshot links in the description are re-pinned to `a84690a` since the old SHA no longer exists on the branch.

**Code fixes** in `2392ffe`: the snapshot is consumed on restore with the dead size guard removed, every notification row is a `label` bound to its switch, All and None start their accessible names with their visible text, a same document fragment is left to the page, and the skeleton draws three cards.

**Tests** in `281acf9`. Thirteen mutations, the seven you listed plus four more for the click handler guards and one for the snapshot, each applied to the source on its own and confirmed to turn the suite red. Four of them survived my first attempt at a test and needed the test reshaped, which is the part your review was right to insist on:

- the plain text click had to move to a separate paragraph, otherwise `view.dom.querySelector('a')` found nothing and the mutation lived
- the raw href mutation needed a relative href, since every earlier case was already absolute
- the button guard needed an actual right and middle click
- the `setPending` mutation only bites when `setPending(false)` moves *inside* the try, not merely out of `finally`, because the catch does not rethrow

The `slack:reaction` seed is back in the round trip test.

**Scope** is now stated in the description under `Scope of the link change`, naming all four surfaces `editorExtensions()` feeds and the double click behaviour, as a decision rather than a side effect.

Lint, comment policy, Bun import policy, dependency dedupe and typecheck are green. The three suites this touches run green directly: 25 in `notification-channels.test.tsx`, 3 in `notification-groups.test.ts`, 12 in `link-click.test.tsx`, and none of them touch the network any more.

One open question, on the fragment thread: an in page link on an editable surface now does nothing rather than scrolling, which matches `main`. Wiring it to `use-hash-scroll` from an editor extension felt like more than this fix should carry, but say the word if you would rather the editable surface matched the read view.

Deferred as you asked, not in this PR: a channel switched off storing one row per currently known type so later added types arrive enabled, nothing in the delivery code reading the `push` channel, and `Collapsible` animating `grid-template-rows` against the motion rule.

apps/web/src/features/docs/editor/link-click.ts
```diff
@@ -0,0 +1,51 @@
...
+    return false;
+  }
+}
+
+export function openableHref(href: string | null, base: string): string | null {
+  if (href === null || href.trim().length === 0) return null;
+  try {
+    const url = new URL(href, base);
+    if (!OPENABLE_PROTOCOLS.has(url.protocol)) return null;
+    if (url.hash.length > 0 && sameDocument(url, base)) return null;
```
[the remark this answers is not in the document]
Tanzeelsameen: Correct, and reachable. Fixed in `2eb1b23`.

Verified the premise before changing anything, because it depends on how `URL` reports an empty fragment:

```
"#"                                  -> https://orbit.test/issues/AM-215#      | hash: ""
"?x=1#"                              -> https://orbit.test/issues/AM-215?x=1#  | hash: ""
"https://orbit.test/issues/AM-215#"  -> https://orbit.test/issues/AM-215#      | hash: ""
```

All three have an empty `hash`, so the guard let them through and the handler opened a second tab of the page the user is already on. Taking your fix as proposed:

```ts
if (href.includes('#') && sameDocument(url, base)) return null;
```

Checked it does not over-reject: a literal `#` in an href always starts the fragment, since one inside a query has to be percent encoded, and `sameDocument` still lets `https://example.com/a#` through because that is another document.

Three regression cases added, `'#'`, the fully spelled out `${base}#`, and `'?tab=activity#'` against a base that already carries that query. Restoring `url.hash.length > 0` turns that test red.


---

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,419 @@
...
+                </Button>
+                <span className="h-5 w-px shrink-0 bg-border" />
+                <Switch
+                  checked={enabled > 0}
+                  onCheckedChange={(next) => setChannelEnabled(channel, next)}
+                  disabled={locked}
+                  aria-label={`${meta.label} notifications`}
+                />
+              </div>
+              <Collapsible open={open}>
```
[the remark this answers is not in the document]
Tanzeelsameen: The reading of the motion rule is right, but this is not this PR's to change, and I am not taking the proposed fix.

`Collapsible` is pre-existing and not touched by this diff. It lives at `apps/web/src/components/ui/collapsible.tsx`, was last modified in an unrelated change, and is used by four other surfaces besides this one: `sidebar-section.tsx`, `sidebar-team.tsx`, `doc-archive-section.tsx` and `doc-tree-rows.tsx`. If animating `grid-template-rows` violates the motion contract, it violates it in all five places, and the fix belongs in the component rather than at one call site.

Swapping this one use for `{open ? ... : null}` would leave the same violation in the other four while making this panel the only expandable surface in the app that appears without transition. That is a worse outcome than the state we are in.

@imshashank raised the same point in his review and explicitly deferred it to be tracked separately rather than fixed here, so this is a decision already taken on this PR, not an oversight.


---

Tanzeelsameen:
@coderabbitai both nitpicks from the summary were valid and are taken in `9748590`.

**`quietHoursEnd` was unasserted.** Correct, and it mattered: the round trip test inspected `quietHoursEnabled`, `quietHoursStart` and `urgentBypassEnabled` but never the end of the window, so the request could stop sending it and the suite stayed green. Confirmed by breaking it, `quietHoursEnd: quietHoursStart` in the payload, which now turns that test red. Assertion added.

**Deep relative imports.** Also correct against `CLAUDE.md`, which says to use `@/` inside `apps/web`. The test tree is split today, 148 files on the alias and 215 on relative paths, so this is not the only offender, but the three suites this branch adds are mine to get right and they now import through `@/`. I applied it to `notification-groups.test.ts` as well, which you did not flag, rather than leave one of the three inconsistent. Converting the other 215 is out of scope here.

Two other things for the record.

The inline finding on `openableHref` and the empty fragment was a real bug and is fixed in `2eb1b23`, with three regression cases. Thanks, that one was a genuine gap in the guard I had just added.

The `Collapsible` finding I have declined, with reasoning on its thread: the component is pre-existing, untouched by this diff, and used by four other surfaces, so taking the proposed fix would leave the same violation in all four while making this the only expandable panel in the app with no transition. @imshashank raised the same point and deferred it to be tracked separately.

Noting for the maintainers: this run reported that it used the open-source allowance rather than the organization plan because the PR author has no assigned seat, and the next included review is 49 minutes out. So a review of the current head may need a seat assigned or a wait.

Tanzeelsameen:
@coderabbitai please do not open an issue for the `Collapsible` transition. @imshashank said he is tracking that one himself, so a second entry would duplicate his.

Tanzeelsameen:
Green and mergeable on `9748590`, so this is now waiting on its reviewers.

All 15 checks pass, including `Unit and integration tests`, `Lint, comments, types`, `Build`, both end to end jobs and both CodeQL analyses. `Deploy documentation` is skipped as usual on a PR. Merge state is `clean` against `main` at `24739ba`, so there is no conflict. Every review thread is resolved with either a fix or a reply.

The description's first checklist item now reads plainly rather than with a caveat, because CI settles it on this head.

**Two review gates are degraded, and neither is something I can clear.**

Greptile has posted `Your trial has ended. Reactivate Greptile to resume code reviews.` on every push, four times now. It has never produced a review of this PR. `CLAUDE.md` names it the primary gate and says its findings carry the most weight, so by the repository's own rule this PR is not ready to merge, whatever the checks say. Reactivating the trial or the subscription needs someone with access to that account.

CodeRabbit has reviewed, and its last run reported: *"This review ran on the open-source allowance, not this organization's plan, because the pull request author doesn't have an assigned seat. Waiting won't change this."* That caps it at one review per hour, which is why its coverage has trailed each push by a commit or two. It has reviewed through `281acf9` and posted three findings, two of which I took and one I declined on its thread; `2eb1b23` and `9748590` are still uncovered, and the next included review is roughly 45 minutes out. Assigning the author a seat would remove the lag.

I will keep watching and re-check periodically. Nothing else is outstanding on my side.

Tanzeelsameen:
@coderabbitai full review

The commit status on the current head, `9748590`, reads `Review rate limited` rather than a real result. `CLAUDE.md` says a rate-limited status is not a review and should be re-run rather than merged on, and your coverage marker still reads `281ac`, so `2eb1b23` and `9748590` are uncovered. The hour window should have reopened by now.

Those two commits are where the empty fragment fix, its regression cases, the `quietHoursEnd` assertion and the `@/` import change landed, which is exactly the work your last review asked for, so it is worth a look rather than a formality.

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,419 @@
...
+    return (value) => {
+      editRevision.current += 1;
+      setSaved(false);
+      set(value);
+    };
+  }
+
+  function toggleType(channel: NotificationChannel, type: NotificationType): void {
+    if (isLocked(channel)) return;
+    apply((next) => {
```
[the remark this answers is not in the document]
Tanzeelsameen: Real, and a genuinely different route into the bug I had just fixed. Taken in `fcca96d`.

I had made `setChannelEnabled` consume the snapshot on restore, which closes the sequence where a master off and on pair happens before the edits. Yours has no intervening master on, so the snapshot from the first master off survives every child edit and is still sitting there when the master is switched on again. Traced it:

1. Load with `email:reaction` disabled. Master off writes the snapshot `{email:reaction}` and disables all 24.
2. Turn Mention on. The channel is now partially on and the master reads on, but `setChannelEnabled(true)` never ran, so the snapshot is untouched.
3. None on all six groups. All 24 disabled, master reads off.
4. Master on restores the step 1 snapshot, so `email:reaction` goes back to disabled and the PUT persists it, although the user's last state through those controls was everything off.

I took the invalidation rather than the conditional version you proposed. `if (disabled.has(key))` in `toggleType` and `if (enabled)` in `setGroupEnabled` cover the paths that re-enable, but the invariant is simpler than that: a snapshot is only meaningful if nothing has touched the channel since the master wrote it. So both helpers now discard it unconditionally:

```ts
if (isLocked(channel)) return;
remembered.current.delete(channel);
```

That is a no-op in every case where no snapshot exists, which is the common one, and it means the snapshot can only ever belong to the off and on pair that created it. Master off, master on with nothing in between still restores exactly as before.

Your sequence is now a test, `forgets a channel snapshot once a type or group control has moved on from it`, walking all four steps and asserting the payload disables nothing. Reverting both `delete` calls turns it red; the earlier restore test and the one for @imshashank's variant both stay green, so the two fixes are independent rather than one masking the other.


---

Tanzeelsameen:
@coderabbitai the Major finding was right and is fixed in `fcca96d`, with the detail on its thread. The remaining nitpick from that review I am declining.

It asks for `@/features/docs/editor/link-click.ts` in place of `./link-click.ts` in `extensions.ts`. That file already imports two siblings relatively:

```ts
import { CodeBlockView } from './code-block-view.tsx';
import { LinkClick } from './link-click.ts';
import { calloutToneOf } from './markdown.ts';
```

Changing only the middle one would make it the single inconsistent import in its own file. Checked the wider convention before deciding: `src` has 495 sibling relative imports, and `src/features/docs/editor/` contains no same directory alias imports at all, so the repository reads the `@/` rule as being about reaching across the tree rather than naming a file next to you.

That is also why the same nitpick was correct on the test files and I took it there. Those were reaching four levels up into `src` with `../../../../`, which is exactly what the alias exists to avoid. A sibling import is not.

Happy to be overruled by a maintainer if the rule is meant literally, but then it is a codemod across 495 imports rather than one line here.

Tanzeelsameen:
@coderabbitai review

The head is `fcca96d`. Your last attempt at it came back rate limited, so the only commit carrying the fix for your own Major finding is still uncovered, and automatic reviews are paused on this branch. The hour window should have reopened.

Two files changed since `9748590`: `notification-channels.tsx` and its test.


---

Tanzeelsameen:
CodeRabbit has now reviewed the current head, `fcca96d`, with **no actionable comments**, and its commit status reads `Review completed` rather than the rate limited result it carried earlier. That closes everything on my side.

State of the PR:

- All 15 CI checks green on `fcca96d`, including `Unit and integration tests`, `Lint, comments, types`, `Build`, both end to end jobs and both CodeQL analyses.
- Merge state `clean` against `main` at `24739ba`. No conflict, and `main` has not moved since this branch forked from it.
- Every review thread resolved, each with a fix or a reply. Four rounds of findings: three from @imshashank and one from CodeRabbit, plus two declined with reasoning, one of which CodeRabbit then withdrew itself.

The one thing still missing is **Greptile**. It has posted `Your trial has ended` on every push, five times, and has never produced a review of this PR. `CLAUDE.md` names it the primary gate whose findings carry the most weight, so by the repository's own rule this is not ready to merge on a clean CodeRabbit alone. Reactivating it needs someone with access to that account, and I have no way to do it from here.

Separately, and worth fixing whoever picks this up: the PR author has no CodeRabbit seat, so every review has run on the open-source allowance at one per hour. That is the whole reason its coverage kept trailing the head by a commit or two through this PR.

I will keep watching and re-check periodically, but nothing further is actionable by me until Greptile is available or a human reviews.

docs/assets/screenshots/notification-settings-customize-dark.png
imshashank, line 1: This is nice.

Tanzeelsameen: Thanks. That is the capture worth looking at, because it is the one carrying the `Switch` change: the off rows are legible against the track only because the unchecked thumb moved to `bg-muted`. On the old `bg-surface` the whole Workspace group would read as an ambiguous smudge in dark mode.

Resolving this one so the thread list stays clean. The PR is green and mergeable on `fcca96d` and waiting on review.


---

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,421 @@
...
+    if (isLocked(channel)) return;
+    const keys = channelKeys(channel);
+    apply((next) => {
+      if (!enabled) {
+        remembered.current.set(channel, new Set(keys.filter((key) => next.has(key))));
+        for (const key of keys) next.add(key);
+        return;
+      }
+      const previous = remembered.current.get(channel);
+      remembered.current.delete(channel);
```
imshashank: This delete sits inside the `setDisabled` updater, so the updater is no longer idempotent. `reactStrictMode` is on in `next.config.ts`, and in development React runs an updater twice and keeps the second result. The first run reads the snapshot and deletes it, the second finds nothing and turns every type on.

It only bites when the updater is not computed eagerly, which is whenever the `setSaved(false)` just before it has real work to do. So the most natural sequence hits it: turn Email off, Save, turn Email back on. I did exactly that in `bun run dev` starting from `On for 23 of 24` and it came back as `On for all 24`. The same sequence in a test wrapped in `<StrictMode>` saves `[]` where it should save `["email:mention"]`. Without StrictMode it restores correctly, so production builds are fine and the suite is green, but the headline behaviour of this PR is broken for anyone running the app locally, and our e2e runs against a dev server.

The placement came from my suggestion in the first review, which put the delete where the `get` already was. That was my mistake. The read and the delete both belong outside the updater, the way you already did it in `toggleType` and `setGroupEnabled`:

```ts
function setChannelEnabled(channel: NotificationChannel, enabled: boolean): void {
  if (isLocked(channel)) return;
  const keys = channelKeys(channel);
  if (!enabled) {
    apply((next) => {
      remembered.current.set(channel, new Set(keys.filter((key) => next.has(key))));
      for (const key of keys) next.add(key);
    });
    return;
  }
  const previous = remembered.current.get(channel);
  remembered.current.delete(channel);
  apply((next) => {
    for (const key of keys) {
      if (previous?.has(key) === true) continue;
      next.delete(key);
    }
  });
}
```

The `set` on the off path can stay where it is, it writes the same value on both runs. I ran this version before suggesting it this time: the StrictMode sequence passes and all 26 of your tests stay green.

Please add the test that would have caught it: render inside `<StrictMode>` with `email:mention` disabled, master off, Save, wait for the saved note, master on, Save again, and assert the second PUT still carries `email:mention` disabled.

apps/web/src/features/settings/notification-channels.tsx
```diff
@@ -0,0 +1,421 @@
...
+      )}
+
+      <div className="flex flex-col gap-3 rounded-lg border border-border bg-surface p-4">
+        <label htmlFor="quiet-hours" className="flex items-center justify-between gap-3">
+          <span className="flex flex-col">
+            <span className="font-medium text-dense text-text">Quiet hours</span>
+            <span className="text-muted text-xs">
+              {props.slackDm === 'available'
+                ? 'Email and Slack DMs are held until the window ends, in your local time.'
+                : 'Email is held until the window ends, in your local time.'}
```
imshashank, line 348: Not blocking and not yours, the wording came over from the matrix. The card says `in your local time`, but the server evaluates the window in `user.timezone`. That column defaults to `UTC`, is `input: false` at sign up, and is only ever set by typing a zone into Settings, Profile. Nothing detects it.

I tested it: with my clock at 16:50 IST and a 16:00 to 18:00 window, the email went out immediately. After setting the profile zone to `Asia/Kolkata`, the same window held it until 12:30 UTC, which is 18:00 IST. So for anyone who has never opened Profile, quiet hours run in UTC and the default 18:00 to 09:00 window is 23:30 to 14:30 for someone in India.

I will track that separately. The real fix is to detect the zone on first sign in, or at least name the zone on this card so the promise is true.

imshashank (commented):
Second pass on `fcca96d`. I re-ran everything rather than reading the replies.

Verified fixed:

- The commits are yours now, there are no trailers, and the tree under the rewritten six is byte identical to what I reviewed the first time.
- The link click test makes zero outbound requests. I checked the probe both ways: it logs the github.com request again the moment I delete your `blockNavigation` line.
- I re-ran the breaks from the first review plus new ones aimed at the fixes themselves, 27 in total, and 23 turn the suite red. Every one that survived last time is caught now. The four survivors: each of the three `remembered.current.delete` calls taken alone, which mask one another (remove them together and two tests fail, so that is redundancy and not a gap), and the skeleton card count, which no test observes and never did.
- 408 tests across the editor, settings and ui suites pass locally. `tsc`, biome, the comment check and the bun import check are clean.
- Live in Chromium against a seeded workspace: clicking anywhere on a row toggles it, All and None move the group count and the channel summary, the quiet hours switch disables both time inputs, and a save of 22:30 to 07:15 with two email picks off lands in the database as exactly that. In a description, `#summary` and a bare `#` open nothing, while an external link and `/issue/ENG-1#notes` open with `noopener,noreferrer`.
- Channel switches do gate delivery, which was the argument for keeping Inbox a real channel. Email off for a type queues no email and still writes the inbox row. Inbox off writes the row with `surface_in_inbox = false` so the inbox never lists it. Both off writes nothing.

One thing left, and part of it is on me: the snapshot fix is not safe under StrictMode. Inline.

I also tested quiet hours end to end since this screen drives it. None of that is your code. Through the real API, with one user assigning issues to another and reading `notification_delivery`:

| Case | Result |
| --- | --- |
| Window contains now, normal issue | email held until the window end, inbox row immediate |
| Same, urgent issue, urgent bypass on | sent immediately |
| Same, urgent issue, urgent bypass off | held |
| Quiet hours off | sent immediately |
| Window does not contain now | sent immediately |
| Overnight window wrapping midnight | held until the window's end time |
| Start equals end | never holds anything |

The cron that drains held mail runs every minute and only claims rows whose `available_at` has passed, so held mail goes out within a minute of the window closing. It all works. There is one gap in what the card promises, inline, not blocking.

Housekeeping: two replies still carry the generated by footer, the skeleton thread and the screenshot thread. Please edit them out.

Fix the updater and I will merge.

apps/web/src/features/settings/notification-channels.tsx
Tanzeelsameen: Confirmed and fixed in `cc2cd6b`. This one is squarely mine: a ref mutation inside a `setDisabled` updater is a side effect in code React requires to be repeatable, and I moved it there without thinking about what runs it.

Reproduced before touching the code. Your sequence as a test, rendered inside `<StrictMode>` with `email:mention` disabled, master off, Save, master on, Save: the second PUT carried

```
SECOND PUT disabled: []
```

where it should carry `["email:mention"]`. Everything on, exactly as you saw in `bun run dev`.

Took your restructure as written. The off path keeps `apply` with the snapshot write inside, since that recomputes the same value from `current` on either pass, and the on path reads and clears in the handler before calling `apply`, so the updater only edits the set it is handed. That also makes all three `remembered.current.delete` calls consistent, which is what I should have done when I added the other two.

`restores the earlier picks under a double invoked state updater` is now in the suite. Reverting the fix turns it red with the `[]` above. All 27 in the file pass, and 71 across the editor, settings and ui suites.

On the placement coming from your first review: your diff put the delete next to the `get`, which was the right instinct, and I moved both without checking that the enclosing function was an updater. The bug is in what I pushed, so it is mine.


---

apps/web/src/features/settings/notification-channels.tsx
Tanzeelsameen: Agreed, and thanks for actually measuring it rather than reasoning about it. The 23:30 to 14:30 example is the part that makes it concrete: the default window is close to useless for anyone who has never opened Profile, and the card claims otherwise.

Leaving it to your separate issue rather than widening this PR. The wording is inherited from `notification-matrix.tsx` and the real fix is on the server side, either detecting the zone at first sign in or writing it from the browser, and naming the zone on the card is only worth doing once the value is trustworthy.

One note for whoever picks it up: `input: false` on the column means the sign up path cannot set it even if the client sent it, so detection needs a deliberate write somewhere after the account exists, not just a wider insert schema.


---

Tanzeelsameen:
@imshashank the blocker is fixed in `cc2cd6b` and pushed. Detail on its thread, short version: reproduced your sequence under `<StrictMode>` and got `[]` where `["email:mention"]` was owed, took your restructure as written, and added `restores the earlier picks under a double invoked state updater` which goes red against the old code. 27 in the file, 71 across the editor, settings and ui suites, lint, comment policy and typecheck all clean.

Thank you for re-running everything instead of reading my replies. The four survivors you found are the right reading of them too: the three `delete` calls genuinely do mask each other, and the skeleton card count is unobserved. That second one is fair and I am not going to pretend otherwise, a card count nobody asserts is a number that can drift.

**On the two footers, I cannot do it and I should have said so earlier rather than letting it reach a second review.** Those lines are appended server side by the tooling after my text is submitted, not typed by me, and the GitHub tooling I have can edit an issue or PR conversation comment but explicitly not a review comment. I cleaned all seven conversation comments once I found the edit worked there. Every inline reply, including the two you named and the two I just posted, is out of reach from here.

So it needs one of you in the web UI, or a token with the same access outside this session. The two you meant:

- the skeleton thread, [`#discussion_r4045708900`](https://github.com/Noveum/orbit/pull/478#discussion_r4045708900)
- the screenshot thread, [`#discussion_r4053035441`](https://github.com/Noveum/orbit/pull/478#discussion_r4053035441)

Worth knowing it will keep happening on anything further I post inline, so if you want the thread list clean before merge it is best done as one pass at the end.

Timezone finding acknowledged on its thread and left to your separate issue. One thing for whoever takes it: `input: false` on `user.timezone` means the sign up path cannot write it even with a wider schema, so detection has to be a deliberate write after the account exists.

CI is re-running on `cc2cd6b`. Base is still `24739ba`, so no merge needed.

imshashank (approved):
Verified `cc2cd6b`.

- The read and the delete now happen in the handler and the updater only edits the set it is given, same shape as `toggleType` and `setGroupEnabled`.
- Your StrictMode test goes red against `fcca96d` and green here. I also moved the delete back inside the updater on this head and it fails, so the behaviour is pinned and not just passing.
- Six breaks of the rewritten function, all caught: restore logic removed, master on made a no-op without a snapshot, the off path not recording a snapshot, the off path not disabling anything, all three deletes removed, and the delete back inside the updater.
- 409 tests across the editor, settings and ui suites pass locally. `tsc`, biome and the comment check are clean.
- In `bun run dev`, the sequence that lost the pick yesterday: `On for 23 of 24`, off, Save, on gives `On for 23 of 24` again, and the second save leaves exactly `email:mention` disabled in the database.

CI is green on this head and every thread is closed. Merging.

One thing to carry into the next PR and not this one: the replies here all end with a generated by footer. Hard rule 3 covers PR text as well as commits, so leave it off from now on. The commits and the description are clean, which is what lands on `main`.

Thanks for turning each round around quickly and for checking every fix against a real break instead of assuming it.

PR #479, opened 2026-09-18
feat(issues): add property undo redo with optimistic cas validation

### LINKED ISSUE #307: [feature]: Cmd+Z does not undo a property change, only text
## Current status, September 22

Implementation is open in #479 by @abdul-rahman-0x. Changes are still requested for two overlapping edits when the later edit succeeds before the earlier edit fails: undo must restore the last saved value, not the failed optimistic value. Keep this issue open until the regression and fresh CI pass.

## What is painful today

Orbit is fast enough that a property gets changed by accident. Press `s`, the status menu opens, an arrow key and Enter land on the wrong row, and the issue is now Done. Or the wrong issue is selected in a list and an assignee is set on it.

At that point `Cmd+Z` is the reflex, and nothing happens. Undo exists inside every text surface, so the reflex is trained: it works in the issue description, in the issue title, and in the comment composer, because TipTap and the browser handle it there. It is only property changes that have no undo, which is the least obvious place for the gap to be.

Recovering means remembering what the value was before. The activity feed records the change, so the old value is on screen, but putting it back is a manual round trip through the same menu that just went wrong. On a keyboard first product this is the one place the keyboard cannot get you out of what the keyboard got you into.

Reproduced on the seeded demo workspace, on `main` at `0c0b301`:

1. Open `/issue/ENG-2`, note the status is In Progress.
2. Click the status property and pick Done.
3. Click on the page background and press `Cmd+Z`.
4. Status is still Done.

Same with estimate, assignee, priority, labels, project and sprint.

## What you would like to happen

`Cmd+Z` undoes the last property change made in this tab, and says so.

The narrow version, which is what I would like to build:

- A small stack of recent property changes, each holding the issue and the patch that puts it back. Capped at something like 20, and living for the session only, not persisted.
- `Cmd+Z` pops the top entry and applies the inverse through the existing `useUpdateIssue` mutation, so it travels the same server path, the same policy check and the same realtime fan out as the original change. An undo is an ordinary write, not a special case.
- A toast confirming what was reverted, so undo is visible rather than silent.
- `Cmd+Shift+Z` to redo the entry just undone.
- The binding registers with `allowInInput: false` so it never competes with TipTap or a native input. Text undo keeps working exactly as it does now, and the two never fight.
- If the server refuses the inverse write, the entry is dropped and the toast says it could not be undone, rather than leaving the stack out of step with the server.

Deliberately out of scope, unless you want them in: undoing a board drag, undoing a create or a delete, and undoing another person's change. Each of those is a different shape, and a stack that mixes them gets hard to reason about quickly. Property changes alone cover the accident that actually happens.

## Which part of Orbit

Issues, lists and boards

## How you work around it now

Read the old value off the activity feed and set it back by hand. It works, and it costs about ten seconds and a small amount of trust every time.

## Anything else

Happy to build this if the shape sounds right, and happy to adjust it first. Two things I would want your call on before writing code:

1. **Session only, or persisted?** I am assuming an in memory stack that dies with the tab. Persisting undo across reloads raises questions about what happens when someone else has changed the issue since, and I would rather not answer those in the first version.
2. **Does undo appear in the activity feed as its own entry, or does it read as a normal change?** I lean towards a normal change, because that is what it is on the server, but it does mean the feed shows a change and then a change back with nothing marking the second as an undo.

Context: I opened [#306](https://github.com/Noveum/orbit/pull/306) for some smaller issue property fixes, and found this while checking whether `Cmd+Z` was broken somewhere. It is not broken anywhere, it just does not exist for actions.

abdul-rahman-0x:
### What this changes
- Extends `issueUpdateSchema` with an optional `expected` precondition schema (`issueExpectedPropertiesSchema`) for optimistic concurrency control.
- Enforces server-side Compare-And-Set (CAS) validation in `packages/core/src/work/issue-service.ts` within the database transaction, preventing undo writes from clobbering concurrent updates from other clients.
- Implements `useIssuePropertyUndo` in `apps/web/src/features/issues/use-issue-property-undo.ts` managing an in-memory session stack capped at 20 entries.
- Binds `mod+z` (undo) and `mod+shift+z` (redo) using Orbit's native `useHotkey` with `allowInInput: false`.
- Wires history recording into property change handlers in `apps/web/src/features/issues/issue-properties.tsx` with dynamic toast confirmations.
- Adds regression test suite in `packages/shared/tests/validators/issue-undo.test.ts`.

### Why
Closes #307

In a keyboard-first application, property modifications happen rapidly with single keystrokes (`s`, `p`, `a`), making accidental edits common. Browser and TipTap undo mechanisms only covered text inputs, leaving property mutations without an undo path.

### How you know it works
- Passed strict Biome linting and formatting across the monorepo.
- Passed zero-comment CI validation gate.
- Passed full TypeScript compilation across all 9 workspaces.
- Passed all 339 unit tests in `@orbit/shared` including new CAS precondition tests.
- Live verified on `localhost:3000`: tested single property undo, redo, and confirmed input isolation in comment composer.

### Screenshots
- Tested live on `localhost:3000`: confirmed toast rendering ("Reverted Status" / "Restored Status") and real-time state reconciliation without TipTap interference.
Reverted Status
Restored Status

### Checklist

### Anything reviewers should know
- The undo/redo history is tab-session only and capped at 20 entries.
- Concurrency conflicts throw `conflict(...)` on the server, which surfaces a danger toast on the client while preserving the remote change.


imshashank (changes requested):
Undo currently overwrites other people's label, reviewer and parent changes because those fields have no expected-value check; I reproduced one member's undo removing another member's label. Please protect these fields inside the locked server transaction and add conflict tests for undo and redo. The history also disappears when IssueProperties unmounts, misses list edits, and can lose earlier rapid edits because recording uses per-call mutation callbacks. Move history to tab lifetime and test navigation, list edits and two edits made before the first request finishes.

abdul-rahman-0x:
Hi @imshashank, I've addressed the requested concurrency and history changes:

- Undo/redo snapshots now always include baseline `labelIds`, `reviewerIds`, and `parentId`, so changes made by another user are detected instead of being overwritten.
- The server validates these expectations inside the locked transaction.
- Added 6 regression tests covering concurrent label, reviewer, and parent changes for both undo and redo. All 6 CAS tests pass.
- Moved property undo/redo history to the workspace/tab lifetime so it is no longer tied to the IssueProperties sidebar lifecycle.
- Full TypeScript typecheck passes across the repository, and Biome passes for the changed files.

The concurrent changes are preserved when a conflict is detected.

imshashank (changes requested):
The server conflict checks and tab lifetime address the earlier findings, but useUpdateIssue now records history in onMutate before the request succeeds. A failed edit clears redo and leaves an undo entry for a change that never happened. I also reproduced a title-only edit producing an undo request containing only expected, with no property to restore. Please record only successful supported property changes, test these cases, and run the full CI workflow on this head.

imshashank (changes requested):
The latest update fixes mixed patches and the simple rapid-edit case. One ordering bug remains: start priority edits 0 to 1 and 1 to 2, let the second succeed, then let the first fail; undo still restores 1 even though that edit never saved. onSettled removes the successful entry from the tracker before the earlier failure can reconcile it. Please retain enough history to handle this order and add a real mutation regression test before merging.

apps/web/src/lib/query/use-issues.ts
```diff
@@ -675,6 +926,31 @@ export function useUpdateIssue() {
...
+        const detailKey = queryKeys.issue(input.issue.identifier);
+        const heldDetail = client.getQueryData<IssueDetail>(detailKey);
+        const fallbackBase = heldDetail?.issue ?? input.issue;
+        tracker = { rootBase: fallbackBase, active: [] };
+        mutationTrackers.set(input.issue.id, tracker);
+      }
+
+      let currentBase = tracker.rootBase;
+      for (const m of tracker.active) {
+        currentBase = applyPatchToIssue(currentBase, m.patch);
```
imshashank, line 940: onError now retains failed records while another mutation is pending, but this loop still applies every record. Start priority 2 to 1 as A, keep an unrelated B pending, let A fail, then start priority 3 as C and let B and C succeed. C records inverse priority 1, so undo restores an edit that never saved. Skip failed records here as reconcileTracker does and cover this sequence with the real hook.

imshashank (changes requested):
The previous response-order fix is present, but a new edit can still inherit a failed mutation. I reproduced this with the real React hook: priority starts at 2, A changes it to 1 and fails while B is pending, then C changes it to 3; undo for C restores 1 instead of 2. Please skip failed records when building currentBase in onMutate and add this ordering as a regression test. The new CI runs are approved and running.

abdul-rahman-0x:
Addressed in 996c780. Failed mutations are now skipped when rebuilding currentBase, and I added a real hook regression test for the A-fails → B-pending → C-succeeds ordering. Focused tests, Biome, and full typecheck pass.

packages/shared/src/validators/issue.ts
```diff
@@ -47,24 +47,41 @@ export const issueCreateSchema = z.object({
...
-    projectId: idSchema.nullable(),
-    milestoneId: idSchema.nullable(),
-    cycleId: idSchema.nullable(),
-    parentId: idSchema.nullable(),
-    estimate: z.number().int().min(0).max(100).nullable(),
-    dueDate: calendarDateSchema.nullable(),
-    labelIds: z.array(idSchema).max(50),
-    sortOrder: z.number(),
+    title: titleSchema.optional(),
+    description: z.string().max(ISSUE_DESCRIPTION_MAX_LENGTH).optional(),
```
imshashank: Please preserve issueDescriptionSchema here. The current head fails apps/web/tests/app/api/issues/[id]/route.test.ts:71: an oversized description returns "Too big: expected string to have <=500000 characters" instead of "Description must be 500,000 characters or fewer." Reuse issueDescriptionSchema.optional() and rerun the route test and full CI.

imshashank (changes requested):
The failed-edit undo bug is fixed, and the exact regression from the last review now passes. One existing API test still fails because issueUpdateSchema replaced issueDescriptionSchema with a plain string limit, losing the description error message. Please use issueDescriptionSchema.optional(), merge current main and run fresh CI, then resolve the remaining review thread. This is the property-undo implementation for #307; coordinate with #492 so we do not ship two histories and competing shortcuts.

apps/web/src/lib/query/use-issues.ts
imshashank: Confirmed fixed in 996c780. The new failed-record filter is present, and both the added test and my independent A-fails/B-pending/C-succeeds reproduction now restore priority 2 correctly. Resolving this finding; the separate description-validation failure still needs fixing.

abdul-rahman-0x:
@imshashank I’ve pushed the requested fix in 84467af and merged the latest main. The 3 workflows are currently awaiting maintainer approval, so fresh CI hasn’t started yet.

Issue #394, opened 2026-09-03
Operations: prove backup, restore, upgrade, and disaster recovery for self-hosted Orbit


imshashank:
## Current status, September 22

Backup capture shipped in #444; guarded restore is being reviewed in #481. Restore still needs authenticated/TLS Redis handling, credential-safe errors and recovery from an interrupted restore. Encryption, retention and the previous-release destroy-and-restore drill remain later slices, so neither PR closes this issue.

## Goal

Turn Orbit’s current backup advice into a reproducible, automated recovery contract covering Postgres, object storage, configuration, and release upgrades.

This is a sibling workstream under #388. It covers the remaining recovery and operator-documentation portions of DB-001, REL-001, REL-002, and DOC-001 in `docs/open-source-readiness.md`.

## Current boundary

The current self-hosting guide says:

- durable application state lives in Postgres and uploaded files live in object storage
- Redis is disposable
- managed Postgres providers may take automatic backups
- a self-hoster should run `pg_dump` and restore it somewhere once

That is sound advice, but it is not yet a supported product contract. There is no upstream backup command, coordinated snapshot procedure, restore validator, scheduled recovery test, retention model, or proof that a database dump and object-store copy describe the same recoverable installation.

## Recovery model

Document and enforce the data classes before writing scripts.

### Durable and required

- Postgres application data, including auth, memberships, workspaces, issues, docs, integrations, MCP grants, and migration ledger
- object-storage uploads and versions required by the supported deletion/recovery policy
- installation/bootstrap state introduced by #393
- operator-supplied configuration that cannot be regenerated safely

### Regenerable or disposable

- Redis fan-out and ephemeral coordination state
- application images, which must be recoverable from immutable release digests
- static assets contained in images
- caches and temporary build/runtime files

Do not back up raw ephemeral access tokens merely because a directory exists. Define every included table/object/config value deliberately.

### 1. Upstream backup command

Provide one supported command or container entry point, for example:

```text
orbit backup create
```

or a documented equivalent under `scripts/`/`deploy/`.

It must:

- verify the installation is on a supported migration state
- record Orbit version, source revision, image digests, migration hashes, database version, and backup format version
- create a consistent Postgres backup using a format suitable for verified restore
- copy or snapshot the object-storage bucket in a way that preserves every object needed by application rows
- include only the minimal configuration metadata needed to restore, never plaintext runtime secrets by default
- produce a manifest containing file/object hashes and counts
- write to an operator-selected destination
[...]

PR #481, opened 2026-09-19
feat(backup): guarded restore engine, validator, and readiness gating

Pallavikumarimdb:
## What this changes

Adds a guarded disaster recovery engine, an application-aware validator, and readiness health gating for self-hosted Orbit. It introduces `bun run backup:restore` and `bun run backup:validate` commands covering Postgres, object storage reconciliation, pre-flight checksum verification, and target database protection.

## Why

Self-hosters currently have manual backup recommendations without an automated or verifiable recovery path. This establishes a reproducible recovery contract covering database restoration, storage synchronization, schema integrity verification, and downtime signalling.

Part of #394

## How you know it works

- Added 40 unit and integration tests covering checksum verification, format compatibility, target confirmation guards, pg_restore pipelines, storage reconciliation, and application validation:
  - `bun test packages/services/tests/backup/`
  - `bun test packages/shared/tests/validators/restore.test.ts`
  - `bun test apps/web/tests/app/api/health/route.test.ts`
- Verified `/api/health` returns 503 Service Unavailable with maintenance metadata while restore or validation is active, and reverts to 200 once complete.
- Tested failure guards: mismatched manifest checksums abort before mutating the target database, and restore aborts if the destination database contains data and confirmation is omitted.

## Checklist

- Target guard: Destructive restores require an explicit `--confirm-target-overwrite` flag or interactive confirmation if existing tables are detected in the destination database.
- S3 reconciliation: Object storage restoration synchronizes missing or mismatched objects by computing SHA-256 digests in-stream, and can be bypassed via `--skip-object-restore` if storage is handled out-of-band.
- Readiness coordinator: Restores set an atomic lock and flag file so reverse proxies and load balancers hitting `/api/health` automatically withhold traffic while tables and objects are in flight.


imshashank (changes requested):
The readiness gate is still unsafe: restoring a backup can replace the current restoring marker with an old ready marker before storage restore and validation finish. Returning 200 when the database check fails also lets the gate fail open, and the status upsert does not prevent two restores running together. Keep maintenance state outside the restored data, acquire exclusive restore ownership, and separate liveness from readiness if the build smoke needs it. Please add a real dump/restore test that checks readiness throughout recovery, including a concurrent second attempt.

packages/services/src/backup/restore.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const compatibility = verifyBackupCompatibility(manifest, options.migrationsFolder);
+
+  await verifyPreMutationChecksums(backupDir, manifest);
+  await setRecoveryState(databaseUrl, 'restoring');
+
+  const driver = resolveStorageDriver(options, env);
+
+  try {
+    await runDatabaseAndMigrations(
```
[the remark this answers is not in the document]
Pallavikumarimdb: [7a8d421](https://github.com/Noveum/orbit/pull/481/commits/7a8d42166748c9f19b7d969817f4161f6615351b)

imshashank (changes requested):
The readiness-table exclusion, exclusive entry, and streaming checksums address the earlier findings. In backup/validate.ts, Redis validation ignores credentials and rediss TLS, so a valid production Redis configuration can leave recovery marked failed; please use the supported Redis client and redact endpoint credentials from errors. In backup/readiness.ts, a killed restore leaves status restoring permanently, so every later restore is rejected even after the original process is gone. Please add authenticated/TLS Redis and interrupted-process recovery tests, implement a safe retry path, and address the open symlink-test thread.

packages/services/src/backup/readiness.ts
```diff
@@ -0,0 +1,134 @@
...
+export interface RestoreLock {
+  readonly release: () => Promise<void>;
+}
+
+export async function acquireRestoreLock(databaseUrl: string): Promise<RestoreLock> {
+  const sql = postgres(databaseUrl, {
+    max: 1,
+    connect_timeout: 5,
+    idle_timeout: 0,
+    prepare: false,
```
imshashank, line 19: postgres.js also has a default max_lifetime of 30 to 60 minutes, separate from idle_timeout. Once that timer closes this idle session, PostgreSQL releases the advisory lock while pg_restore or storage reconciliation may still be running, allowing a second destructive restore. I reproduced this with the real client and database by accelerating only its lifetime timer. Disable automatic lifetime recycling for this lock connection and ensure unexpected connection loss stops the restore safely.

imshashank (changes requested):
The Redis authentication, TLS and redaction fixes pass, and all 55 backup tests pass locally. One production lock issue remains: idle_timeout: 0 does not disable postgres.js max_lifetime, which closes this connection after 30 to 60 minutes. I reproduced a second restore acquiring the lock before the first released it by accelerating that timer. Please keep the lock connection alive for the whole restore, handle unexpected lock loss safely and add a regression.

packages/services/src/backup/restore.ts
```diff
@@ -0,0 +1,174 @@
...
+    let objectsReconciled = 0;
+    if (driver !== undefined) {
+      lock.assertActive();
+      const storageResult = await restoreStorageObjects({
+        objectsDir: join(backupDir, 'objects'),
+        expectedObjects: manifest.checksums.objects,
+        driver,
+      });
+      objectsReconciled = storageResult.uploadedCount + storageResult.verifiedCount;
+    }
```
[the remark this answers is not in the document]
Pallavikumarimdb: [da7f4c5](https://github.com/Noveum/orbit/pull/481/commits/da7f4c579e069f837fa0e797abf2dad151df21f7)

packages/services/src/backup/validate.ts
```diff
@@ -0,0 +1,563 @@
...
+      const objStat = await driver.stat(safeKey);
+      if (objStat === null) {
+        missingObjects += 1;
+        errors.push(
+          `Referenced object "${safeKey}" for attachment "${att.id}" was not found in storage.`,
+        );
+        continue;
+      }
+
+      if (objStat.size !== att.size) {
```
imshashank: Metadata alone does not prove the attachment can be read. After the removed driver.get check, validateRestore returns valid: true for a referenced object with matching stat metadata but no readable body; I reproduced that on this commit. #394 explicitly requires referenced objects to exist and be readable with the expected constraints before readiness becomes green. Restore a bounded body/readability check and a regression for failed reads.

imshashank (changes requested):
The lock lifetime and lock-loss fixes are present, and all 57 backup tests pass. This update also removed the object-read check from restore validation, so matching metadata now passes even when the file cannot be read. I reproduced valid: true with one referenced attachment whose stat succeeds but get returns null, which violates #394 and can reopen an unusable installation. Please restore bounded read and size validation, add that regression, then update main and rerun CI.

packages/services/src/backup/readiness.ts
imshashank: Confirmed fixed in da7f4c5. Automatic connection lifetime recycling is disabled, lock loss aborts the restore, and the new loss tests pass with the full 57-test backup suite. Resolving this finding; the newly removed object-readability check is covered by the new review.

packages/services/src/backup/validate.ts
Pallavikumarimdb: [a527da6](https://github.com/Noveum/orbit/pull/481/commits/a527da662945c2f5d335642b18ffd2ed042fc980)

